PR #3, opened 2015-01-13
Updated to Wagtail 0.8

kaedroho:
Also updated the base box, vagrant configuration and requirements.


davecranwell:
You're as eligible to sense-check this as anyone, so I say just go ahead and merge it

PR #14, opened 2016-02-05
Add a Google ad grants page

mope:
This is the first iteration of the new Google ad grants application page.


tbx/core/static/torchbox/css/main.css
```diff
@@ -2975,3 +2975,328 @@ h4 ~ .stat {
         page-break-after: avoid;
     }
 }
+
```
davecranwell: All this CSS looks like the wrong intent amount (should be 4 space not 2) and it's also pretty horribly formatted either way: closing braces on the same lines, no space between rules etc.

PR #19, opened 2016-03-20
Add marketing landing page

spapas:
- Add new ``Page`` type (same with ``HomePage``) named ``MarketingLandingPage`` with editable related links to be used instead of the hardcoded ones (who we are, get in touch) that the homepage has. The instance of ``MarketingLandingPage`` should be placed as a child of the root (next to blog index and work index) and be visible in menus.
- Keep ``HomePage`` styling by using same class (``page-homepage``) on body - to do this, the content_type tag is changed to return ``homepage`` for the marketing page. Probably a better solution would be to return the correct content type (``marketinglandingpage``) and change the css for the ``page-marketinglandingpage`` class to be similar to to the ``page-homepage`` - however I don't want to mess with the css.
- Copy the ``HomePage`` template to the ``MarketingLandingPage``  - with some small changes
- Add a background jpg (a photo of a foosball table) for marketing landing page
- Add a ``marketing_only`` flag to blog posts and items. When this is set, these objects will be displayed only on the marketing landing page and not on the home page or the corresponding blog/work index.
- Marketing landing page should contain only blog posts and work items tagged by "digital marketing". To do this, we add an optional ``marketing`` flag input to ``work_and_blog_listing`` template tag that filters the returned work items and blog posts by "digital_marketing" if set. The marketing landing page template sets the ``marketing`` flag  to ``work_and_blog_listing`` to return only these items and also to list ``marketing_only`` flagged items (which won't be listed in the homepage).
- Add the clients block to marketing landing page: The clients block is a bunch of links contained in the about page. In order to avoid repeating the same rows from the about page, the clients block of marketing we just get the clients from the about page (``get(slug=about)``) and renders them using a partial template (which is also used in the about page for dryness). Notice that the ideal solution for the clients block would be to use a snippet to model that block. This snippet would then be used to both the about page and the marketing landing page. Adding the snippet of course would mean that the existing client entries would be to re-inserted to the database so I preferred to just re-use the existing client rows from "about".


gasman:
All looks good to me! Merged/deployed to staging now - I've created an example Marketing page at https://tbxwagtail-staging.torchboxapps.com/marketing/

PR #23, opened 2016-04-14
Feature/noticket sign up form page

justinepocock:
Small js and css tweaks


justinepocock:
Right finally found the transition problem. Working now!

PR #25, opened 2016-05-03
Use text and svg for "Made with Wagtail" footer

alexgleason:
Here's a screenshot from my 3K laptop screen:

screenshot from 2016-05-03 15 21 48

It's now selectable text except for "wagtail" which uses an SVG background.


zerolab:
@helenb can you please review/update?

helenb:
Just reviewing the commits and I see there are no scss files here - only css ones. Also since the original screenshot the footer has changed - although this update to use text rather than an image for the wagtail part still applies. It might be easier for me to add these changes in to a new branch by hand.

zerolab:
Yes please

helenb:
Have now extracted and adapted this code in 129-made-with-wagtail-footer which I will pass over to Ashley to approve.

PR #31, opened 2016-08-08
Fix pull_media local dir

kaedroho:
The task was recently changed to use `$CFG_MEDIA_DIR`, which doesn't contain the `/*` on the end of the path meaning that rather than syncing the contents of the folder, it now syncs the folder itself so media files end up in `/vagrant/media/media` instead of `/vagrant/media`

This pull request appends '*' to the media dir again (hopefully) fixing the issue

Note: I haven't tested this (the issue was reported by another torchbox employee)


zerolab:
No longer relevant

PR #45, opened 2016-09-12
Fix js error

zerolab:


nimasmi:
Sorry Toby, that I've not looked at this sooner, and this code has now gone stale. I'm not sure which changes are from yours and other commits in main.js. Can you please merge master into this branch?

Loque-:
I rebased the branch and re-added the change (as it was quicker than trying to resolve the conflict manually) - the conflict was due to the someone very kindly fixing the indentation in the JS file.

Loque-:
The FE assets need to be re-created now this change is in the source, I am going to try and do this on master quickly and commit it there. I have a hunch this won't work due to the git setup.

Loque-:
Sorry for the noise, just improving the fix and generating assets so it can be deployed with less pain.

PR #58, opened 2016-09-15
Feature/syntax highlighting

nimasmi:


nimasmi:
I thought that feature had disappeared off the face of the information superhighway when Django moved from South-based migrations to native 1.7+ migrations.

Thank you for this criticism, as otherwise I wouldn't have known that it was still possible (without renaming files). I completely agree.

https://docs.djangoproject.com/en/1.10/ref/django-admin/#cmdoption-makemigrations--name

nimasmi:
Update: it had. It was [new in Django 1.8](https://docs.djangoproject.com/en/1.9/topics/migrations/#workflow).

zerolab:
Ping @mope

nimasmi:
Ping @mope

tbx/core/wagtail_hooks.py
```diff
@@ -59,5 +59,12 @@ class SubmissionsModelAdminGroup(ModelAdminGroup):
     menu_order = 600
     items = (SignUpFormPageResponseModelAdmin, GoogleAdGrantApplicationModelAdmin)
 
-
 wagtailmodeladmin_register(SubmissionsModelAdminGroup)
+
+
+@hooks.register('insert_global_admin_css')
+def import_fontawesome_stylesheet():
+    elem = '<link rel="stylesheet" href="{}torchbox/vendor/fontawesome/css/font-awesome.min.css">'.format(
```
m1kola: It's sad that we need  whole font awesome only for `code` icon. Is that so important? Could we use `cogs` instead? :)

m1kola: if we need font awesome, would be better to use this instead:

```
from django.contrib.staticfiles.storage import staticfiles_storage
return staticfiles_storage.url(path)
```

So we will be able to use non-default storage

requirements.txt
```diff
@@ -20,3 +20,7 @@ ConcurrentLogHandler==0.9.1
 
 # Fix for html5lib issue, see https://github.com/torchbox/wagtail/issues/2843, fixed in Wagtail 1.4.6 and 1.5.3
 html5lib==0.999999
+
+# Wagtail-markdown
+git+https://github.com/torchbox/wagtail-markdown.git@mediafix#egg=wagtail-markdown
```
m1kola: Can we merge `mediafix` into the `master` and create a package in pypi? :)

BTW, there is a request for package: https://github.com/torchbox/wagtail-markdown/issues/12

zerolab: @m1kola do you want to take lead on that?

re: wagtail-markdown dependency -- are you proposing anything other than the `mediafix` merge and package release?

tbx/core/wagtail_hooks.py
zerolab: Font Awesome is used for the simplemde editor icons, so not only the `code` icon specifically used for the new markdown block.

requirements.txt
m1kola: > @m1kola do you want to take lead on that?

I don't know a reason why `mediafix` have not been merged into `master`, but I can help with a package. Actually it's pretty simple: we can just copy and paste `MANIFEST.in`, `setup.cfg` and `setup.py` form the [wagtailsurveys](https://github.com/torchbox/wagtailsurveys) repo ;)

> are you proposing anything other than the mediafix merge and package release?

No, we can simply merge `mediafix` into `master` and release the package.

zerolab: I can't see any other reason than Liss being swamped with other work.
https://github.com/torchbox/wagtail-markdown/compare/mediafix looks straightforward and we've been using on the intranet for some time.

zerolab:
Nudge

nimasmi:
I got tied up with the prospect of yak shaving wagtailmarkdown. I'm still happy to merge this without that.

zerolab:
Let's merge it in as it is, then once the yak shaving is done we can amend

PR #68, opened 2016-11-03
Reduce size of blog thumbnails on the homepage

kaedroho:
Currently, some images in the blog feed are over 2MB, but they are being displayed much smaller than their actual size.


kaedroho:
On staging

PR #76, opened 2016-11-29
Keep pip up to date and don't run migrate on initial build

kaedroho:
Updating pip keeps away annoying warning messages and probably makes installs faster.

Running migrate on initial build is pointless unless you plan on writing content from scratch.


zerolab:
LGTM

nimasmi:
The vagrant imports line is worth backporting elsewhere too.

zerolab:
Updated as per latest wagtail-kit

PR #83, opened 2017-01-17
Wagtail upgrade

m1kola:


nimasmi:
There's a use of `|embed` at https://github.com/torchbox/wagtail-torchbox/blob/fd62b38f461092484632dcae2e5ef7d3f47e1f94/tbx/core/templates/torchbox/google_ad_grants_page.html#L54, which should be changed, as per http://wagtail.readthedocs.io/en/v1.8/releases/1.7.html#embed-template-filter-is-now-a-template-tag (whether that template is used or not, I don't know).

m1kola:
Hi Nick, thanks for the review, but I've already done it in https://github.com/torchbox/wagtail-torchbox/commit/4a08308a60a0db439a31481384ee7bc61462622b#diff-300fbada2e258d6b8d9a5fbc948c7bf1R54 ;)

I'm going to rebase this branch on top of the `master` branch and I think it will be ready for deploy.

nimasmi:
Good o.

kaedroho:
Looks like the search backend needs to be replaced with ``wagtail.wagtailsearch.backends.elasticsearch``.

https://sentry.torchbox.com/torchbox/torchbox-website/issues/4830/

kaedroho:
Also, this is probably an issue caused by this PR as well: https://sentry.torchbox.com/torchbox/torchbox-website/issues/4831/

m1kola:
I've fixed search issue in 3ca62bce4595de78810eb7f5501e227af856bd9d. Trying to find out what happened with the `WorkPage` model.

m1kola:
Second issue was fixed in 5815bcbdc9758eceeecfa2eb49761ab7d4e28044.

This issues was caused by this PR, but introduced earlier. It was just a wrong call of `PageChooserBlock`: 

```
PageChooserBlock(['torchbox.WorkPage'])
```

Wagtail 1.6.2 didn't allow to pass a target model into `PageChooserBlock`. So the `['torchbox.WorkPage']` list was silently used as `can_choose_root` argument, but in Wagtail 1.8 the signature of the `PageChooserBlock` class has been changed. Now it accepts `target_model` as the first argument and `can_choose_root` as the second. But `target_model` can't be a list anyway.

Refs:
* https://github.com/wagtail/wagtail/blob/v1.6.2/wagtail/wagtailcore/blocks/field_block.py#L530
* https://github.com/wagtail/wagtail/blob/v1.8/wagtail/wagtailcore/blocks/field_block.py#L576

PR #107, opened 2017-10-12
Revert "Rewrites WorkPage with a brand new design"

kaedroho:
Reverts torchbox/wagtail-torchbox#97


kaedroho:
Need to make the migrations reversible first.

PR #110, opened 2018-03-20
Wagtail 2.1.1

nimasmi:
Python 3.6, Django 1.11 and Wagtail 2.1.1

Switch to Postgres search backend.
Other fixes to make things work.


requirements.txt
```diff
@@ -1,12 +1,10 @@
-Django==1.9.12
-psycopg2==2.6.1
-elasticsearch==1.2.0
+Django==1.11.11
+psycopg2==2.7.4
 django-redis==4.3.0
-wagtail==1.9
+wagtail==2.0
```
zerolab: Why not bump this to 2.1 while at it?

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+ignore = E501,F405
```
m1kola: We shouldn't probably ignore `E501` (line len) and `F405` (star imports):

* We need to set 120 char limit
* Start imports are explicitly ignored by the `# noqa` comment

tm-kn:
Moving to #124

PR #112, opened 2018-04-16
217 add captcha to contact form

bcdickinson:


tbx/settings/local.py.example
```diff
@@ -9,6 +9,9 @@ GOOGLE_MAPS_KEY = 'get-one-from-https://code.google.com/apis/console/?noredirect
 # in your version control system.
 SECRET_KEY = 'enter-a-long-unguessable-string-here'
 
+# Get a key for localhost here: https://www.google.com/recaptcha/admin
```
tomdyson: I think we should use environment variables locally as well, e.g. with https://github.com/jpadilla/django-dotenv

PR #118, opened 2018-06-22
WIP: 215 add links to streamfield blocks

jafacakes2011:
> I need to be able to link to external links or internal pages in all streamfield blocks within the service pages - **Will H**


jafacakes2011:
I have tested this change locally and it adds the needed functionality, the only issue I can see is that the Rich text block looks strange for an editor.

screen shot 2018-06-22 at 16 32 50

richbrennan:
LGTM

jafacakes2011:
Can this be merged to staging so it can be reviewed?

martycoote:
FE changes waiting for code review.

helenb:
Shame we have to duplicate the styles, but if there is no way to add a class to the rich-text div then we'll have to live with that. LGTM.

PR #119, opened 2018-06-28
WIP: Updating to continue using hallo.js instead of draftail after upgradi…

victoriachan:
I've updated the hallo.js plugin according to the documentation. I haven't updated `span.js` file as I am assuming it remains unchanged. I haven't tested this yet as I can't get a local build in 2.0 to work. 

Adding WIP as I haven't tested it yet, but feel free to remove WIP and merge if tested OK.

### DIFF
--- a/tbx/core/wagtail_hooks.py
+++ b/tbx/core/wagtail_hooks.py
@@ -7,22 +7,17 @@ from wagtail.contrib.modeladmin.options import ModelAdminGroup, ModelAdmin, mode
 
 from .models import GoogleAdGrantApplication, SignUpFormPageResponse
 
+from wagtail.admin.rich_text import HalloPlugin
 
-@hooks.register('insert_editor_js')
-def editor_js():
-    js_files = [
-        'torchbox/js/hallo-plugins/span.js'
-    ]
-    js_includes = format_html_join(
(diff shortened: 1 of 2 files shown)


tbx/settings/base.py
```diff
@@ -193,6 +193,15 @@
 }
 
 
+# Use hallo.js editor instead of draftail after upgrading to 2.0
+
+WAGTAILADMIN_RICH_TEXT_EDITORS = {
+    'default': {
+        'WIDGET': 'wagtail.admin.rich_text.HalloRichTextArea'
```
m1kola: It's not a good idea to continue using hallo, because it's deprecated. It will not receive any bug fixes, etc. Also it will be removed from Wagtail at some point.

tbx/core/wagtail_hooks.py
```diff
@@ -7,22 +7,17 @@
...
-        </script>
-        """
+
+@hooks.register('register_rich_text_features')
+def register_embed_feature(features):
+    features.register_editor_plugin(
+        'hallo', 'span',
+        HalloPlugin(
+            name='spanbutton',
+            js=['torchbox/js/hallo-plugins/span.js'],
```
m1kola: I didn't manage to find any places where we use it. So I'm going to remove it.

PR #120, opened 2018-07-02
Dokku changes

jafacakes2011:
Trying to make the project Dokku compatible


Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    apt-get install -y nodejs &&\
+    rm -rf /var/lib/apt/lists/*
+
+# Install Gunicorn.
+RUN pip install "gunicorn>=19.8,<19.9"
+
+WORKDIR tbx/static_src
+
+# Install front-end dependencies.
+COPY ./tbx/static_src/package.json ./tbx/static_src/package-lock.json ./
```
tm-kn: We don't use NPM for this project since it's old. We use something called Django compressor. You need to take out the  NPM specific bits. You just need to use `django-admin compress`, example Dockerfile:

```Dockerfile
# Build Python app.
FROM python:2.7-stretch

WORKDIR /app

ENV PYTHONUNBUFFERED=1 \
    PYTHONPATH=/app \
    DJANGO_SETTINGS_MODULE=tbx.settings.production \
    PORT=8000 \
    WEB_CONCURRENCY=3 \
    GUNICORN_CMD_ARGS="--max-requests 1200 --access-logfile -"

EXPOSE 8000

# Install operating system dependencies.
RUN apt-get update -y && \
    apt-get install -y apt-transport-https rsync && \
    rm -rf /var/lib/apt/lists/*

# Install Gunicorn.
RUN pip install "gunicorn>=19.8,<19.9"

# Install Python requirements.
COPY requirements.txt /
RUN pip install -r /requirements.txt

# Install application code.
COPY . .

# Don't use the root user.
RUN useradd tbx
RUN chown -R tbx .
USER tbx

# Install assets
RUN SECRET_KEY=none django-admin collectstatic --noinput --clear

# Compress
RUN SECRET_KEY=none django-admin compress

# Run application
CMD gunicorn tbx.wsgi:application
```

Look at the current deployment process to understand what happens:

https://github.com/torchbox/wagtail-torchbox/blob/be602169c1a12a14bdbc7d2d26f879c7c1097db9/fabfile.py#L41-L59

heroku.yml
```diff
@@ -0,0 +1,8 @@
+build:
+  docker:
+    web: Dockerfile
+release:
+  image: web
+  command:
+    - django-admin createcachetable
+    - django-admin migrate --noinput
```
tm-kn: Probably worth configuring your editor to add a newline at the end of the file.

tbx/settings/base.py
```diff
@@ -167,14 +179,12 @@
...
 # http://django-compressor.readthedocs.org/en/latest/settings/
-
 COMPRESS_PRECOMPILERS = [
     ('text/x-scss', 'django_libsass.SassCompiler'),
 ]
 
 
 # Use Redis as the cache backend for extra performance
-
 CACHES = {
```
tm-kn: Since you specify `CACHES` settings below this is redundant. We want to use Redis on Heroku and database on Dokku since it's not crucial for this project.

tbx/settings/base.py
```diff
@@ -186,19 +196,160 @@
...
+            'LOCATION': 'database_cache',
+        }
+    }
 
-# Use Elasticsearch as the search backend for extra performance and better search results
 
+# Use Elasticsearch as the search backend for extra performance and better search results
 WAGTAILSEARCH_BACKENDS = {
     'default': {
         'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch',
```
tm-kn: Probably worth changing to Postgres backend since this website does not need very good search. ES is expensive to run. Postgres is free.

http://docs.wagtail.io/en/v2.0/reference/contrib/postgres_search.html

Also please do delete elasticsearch dependencies in requirements.txt if you decide not to use ES.

heroku.yml
tm-kn: Since you use VS Code, I think the setting is `"files.insertFinalNewline": false,`

tbx/settings/base.py
```diff
@@ -186,19 +196,160 @@
...
+            'class': 'raven.contrib.django.raven_compat.handlers.SentryHandler',
+        },
+    },
+    'formatters': {
+        'verbose': {
+            'format': '[%(asctime)s][%(process)d][%(levelname)s][%(name)s] %(message)s'
+        }
+    },
+    'loggers': {
+        '{{ cookiecutter.repo_name }}': {
```
tm-kn: Change it to the top level module, i.e. `tbx`. That means any log from that module or submodule will use this logger.

tbx/settings/base.py
```diff
@@ -186,19 +196,160 @@
...
+
+if env.get('SECURE_BROWSER_XSS_FILTER', 'true').lower().strip() == 'true':
+    SECURE_BROWSER_XSS_FILTER = True
+
+if env.get('SECURE_CONTENT_TYPE_NOSNIFF', 'true').lower().strip() == 'true':
+    SECURE_CONTENT_TYPE_NOSNIFF = True
+
+# Referrer-policy header settings
+# https://django-referrer-policy.readthedocs.io/en/1.0/
+REFERRER_POLICY = env.get('SECURE_REFERRER_POLICY',
```
tm-kn: This needs adding that plugin to the requirements.txt file.

tbx/settings/base.py
```diff
@@ -1,21 +1,36 @@
...
+DEBUG = False
+
+if 'SECRET_KEY' in env:
+    SECRET_KEY = env['SECRET_KEY']
+
+if 'ALLOWED_HOSTS' in env:
+    ALLOWED_HOSTS = env['ALLOWED_HOSTS'].split(',')
+
 
 ADMINS = (
```
tm-kn: We do not need this setting if we just use Sentry, see https://docs.djangoproject.com/en/2.0/topics/logging/#django.utils.log.AdminEmailHandler

tbx/settings/base.py
```diff
@@ -1,21 +1,36 @@
...
     ('Nick Smith', 'nicks@torchbox.com'),
 )
 
 MANAGERS = ADMINS
```
tm-kn: same as ADMINS.

tbx/settings/base.py
```diff
@@ -1,21 +1,36 @@
...
+        env[key[4:]] = value
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 BASE_DIR = os.path.dirname(PROJECT_DIR)
 
+# Basic settings
+DEBUG = False
+
+if 'SECRET_KEY' in env:
```
tm-kn: This settings is duplicated in production.py. Make sure other settings are not duplicated before the next review. :P

tbx/settings/base.py
```diff
@@ -186,19 +196,160 @@
...
+    if 'AWS_S3_CUSTOM_DOMAIN' in env:
+        AWS_S3_CUSTOM_DOMAIN = env['AWS_S3_CUSTOM_DOMAIN']
+
+    if 'AWS_S3_SECURE_URLS' in env:
+        AWS_S3_SECURE_URLS = (
+            env['AWS_S3_SECURE_URLS'].strip().lower() == 'true'
+        )
+
+    INSTALLED_APPS += (
+        'storages',
```
tm-kn: Make sure django-storages and boto3 are in requirements.txt in order to use this.

tbx/settings/base.py
tm-kn: Wait... Since it's an old version of Wagtail, it's worth not to do that...

tbx/settings/base.py
jafacakes2011: Where would the best place for this be, base or production?

tm-kn: If it does not have to be in production, then probably worth having it in the base config, so you get config that is the same on dev environment (easier to sport syntax and whatnot errors).

Dockerfile
m1kola: It doesn't look like we use Django compressor anymore: At least, I can't find any `compress` template tags in use. Probably someone who was removing it from the project didn't remove all the traces.

Also front-enders work on new tooling in #122 which is not quite ready... Not sure if we should wait for tooling to be ready or not. Will have a look at this today.

tm-kn:
Taking this over.

tm-kn:
Moved this to #124. Thanks Cameron!

PR #125, opened 2018-08-10
Removes npm scripts from initial provisioning

m1kola:
Mac OS users are having issues running `npm install` within the VM across the shared folders, so this PR:

* Removes `npm` commands from provision scripts to speed up first `vagrant up` for Mac OS 
* Adds docs on how to install required version of `nodejs` on the host machine

We will still have `nodejs` installed in the VM, so users will have a choice where to run npm: inside the VM or on the host machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 .idea/
 node_modules*
 /None
+/npm-debug.log
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/argon
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ We recommend running Wagtail in a virtual machine using Vagrant, to ensure that 
 ### Installation
 
 Run the following commands:
-	
-```
+
+```bash
 git clone [the url you copied above]
 cd wagtail-torchbox
 vagrant up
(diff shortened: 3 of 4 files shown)


tm-kn:
Btw, the same happens on Linux.

helenb:
If this is merged, we probably no longer need the changes to tooling added for https://git.torchbox.com/internal/wagtail-kit/issues/169

It would speed up the tooling to remove that.

m1kola:
> Btw, the same happens on Linux. It's probably to do with VM memory.

There are two issues:

* npm install and scripts are slow on Mac OS. This PR solves the issue by giving Mac OS users a choice where to run npm: in slow VM or on the host
* npm fails in VM. This is, probably, because of memory. Vagrant creates VMs with 512 MiB of memory by default. I tried to extend it to 1024, but it did fail anyway. This PR will help users to avoid this issue by running npm on host. I'll try to upgrade nodejs to the latest LTS version separately to see if it uses less memory.

helenb:
See my update above - @m1kola do you have time to revert the commits associated with that issue as part of this update?

tm-kn:
@helenb I think the issue is open-ended since Mikalai will test with newer Node. Not sure if it's worth the effort now.

helenb:
@m1kola @tm-kn what came of this? Can we remove the changes we added in https://git.torchbox.com/internal/wagtail-kit/issues/169? (there we added a change to rebuild node-sass every time to prevent issues between installing on vm / installing on host).

PR #122, opened 2018-07-03
221 frontend tooling

hippogriffic:
This needs testing on staging before it gets merged

updates frontend tooling to the latest set up from wagtail kit
- removes gulp in favour of npm scripts
- uses node-sass, webpack

needs backend to look in the right place for static files

then needs testing to make sure compiled files are working correctly

PR #126, opened 2018-08-10
Upgrades nodejs to the latest LTS version

m1kola:
In #125 we've removed `npm` calls from vagrant VM provisions, because of performance issues on Mac OS.

There is another issue: `npm` gets killed inside of the VM on `npm install`. This seems to happen, because it uses a lot of memory.

This PR upgrades nodejs from version 4 to 8 (it is currently the latest LTS version). Now it is possible to run npm inside of VM with 512 MiB of memory (I tested it several times on Mac OS). It is still slow on Mac OS, it is better now. Probably, we should still have `npm install` and `npm run build:prod` to be manual steps.

Another advantage of the upgrade: we now have `package-lock.json`.

I didn't upgrade dependencies, because they will be changed as result of #122

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/argon
+lts/carbon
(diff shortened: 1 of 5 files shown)


tm-kn:
Merged as part of #122

PR #130, opened 2018-08-29
Updated Readme after trying out a new build

victoriachan:
I've updated the Readme after testing it out locally. Please check it makes sense and commit. Thanks.

In particular, I needed to run `dj createcachetable` on my build as it was complaining about missing cache table. Just want to double check we do have to do it manually on new VM, and it's not something that the provisioning should do for you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,14 +23,18 @@ git clone [the url you copied above]
 cd wagtail-torchbox
 vagrant up
 vagrant ssh
+
 # then, within the SSH session:
-./manage.py createsuperuser
-./manage.py runserver 0.0.0.0:8000
+
+dj createcachetable
+dj createsuperuser
+djrun
 ```
 
 To build static files you will additionally need to run the following commands:
 
 ```bash
+cd /vagrant/tbx/core/static_src/
 npm install
 npm run build:prod
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,14 +23,18 @@ git clone [the url you copied above]
 cd wagtail-torchbox
 vagrant up
 vagrant ssh
+
 # then, within the SSH session:
-./manage.py createsuperuser
-./manage.py runserver 0.0.0.0:8000
+
+dj createcachetable
+dj createsuperuser
```
richbrennan: I think you'll need to run `dj migrate` if its a fresh build

victoriachan: Thanks. Added `dj migrate` now.

README.md
```diff
@@ -23,14 +23,19 @@ git clone [the url you copied above]
 cd wagtail-torchbox
 vagrant up
 vagrant ssh
+
 # then, within the SSH session:
-./manage.py createsuperuser
-./manage.py runserver 0.0.0.0:8000
+
+dj createcachetable
```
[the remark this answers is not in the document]
victoriachan: Hi Tom, I've corrected this now.

PR #132, opened 2018-08-30
216 staging

victoriachan:
This is a duplicate of `216-add-paragraph-to-service-page` block to be merged to staging. The original branch had migration conflict with another undeployed ticket (`215-add-links-to-streamfield-blocks`) on staging.


victoriachan:
Oh actually will delete this.

PR #135, opened 2018-09-06
WIP: 227 Make newsletter email signup field mandatory

victoriachan:
See https://projects.torchbox.com/projects/torchboxcom/tickets/227


richbrennan:
Do we know which browsers this won't work for, and whether they generate much traffic to the site?

tomdyson:
Only ancient browsers don't support it. It's not a problem for our audience.

https://www.w3schools.com/tags/att_input_required.asp

victoriachan:
Great, thanks.

PR #137, opened 2018-09-18
set a min height on hero and image to prevent text overlap

hippogriffic:
up on staging, ready for review


helenb:
That works fine, thanks Beth. Happy for that to be merged.

PR #138, opened 2018-09-28
Add feed_image to WorkPage

tm-kn:


tbx/core/models.py
```diff
@@ -1043,6 +1043,13 @@ class WorkPage(Page):
         on_delete=models.SET_NULL,
         related_name='+'
     )
+    feed_image = models.ForeignKey(
```
tomdyson: `help_text`? I wouldn't know what this was for.

tm-kn: Do you think we shall add help_text for every other `feed_image` field as well? :stuck_out_tongue:

gasman:
Merged via #145

PR #139, opened 2018-09-28
Fix Planet Drupal feed image URLs

tm-kn:


zerolab:
We no longer need the Planet Drupal feed, so you are better of removing it

PR #141, opened 2018-10-03
Prevent double-escaping HTML entities in Twitter description

gasman:
The `|striptags` filter leaves HTML entities such as `&#x27;` unchanged, and doesn't mark its output as HTML-safe, so the output ends up being double-escaped as `&amp;#27;`. This causes some consumers (specifically Slack) to display the entity in escaped form.

The double-escaping can be seen on this blog post: https://torchbox.com/blog/were-supporting-google-nasa-oxfam-nhs-wagtail-now-we-can-support-you-too/

([the Django docs](https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#striptags) explicitly say not to do this, but `self.intro` is as trusted as any other bit of rich text content in Wagtail, so I believe it's a legitimate thing to do here...)


zerolab:
LGTM. @tombola can you sort this out as well?

PR #144, opened 2018-10-15
Upgrade to Wagtail 2.2 and install wagtail-review

gasman:


zerolab:
@tombola everything apart from requests was tested on staging. I suggest deploying this updated work to staging, give it a sense check, then let's get it live!

PR #151, opened 2019-01-16
Restructure

kaedroho:
The "core" app is starting to get a bit bloated. This moves the services, work and blog models into their own apps.


tbx/work/models.py
```diff
@@ -0,0 +1,183 @@
...
+
+    content_panels = [
+        FieldPanel('title', classname="full title"),
+        FieldPanel('intro', classname="full"),
+        FieldPanel('hide_popular_tags'),
+    ]
+
+    promote_panels = [
+        MultiFieldPanel(Page.promote_panels, "Common page configuration"),
+        FieldPanel('show_in_play_menu'),
```
zerolab: Since you are in the area, it would probably be good to check and get rid of all "Play" bits of code.
It is cruft we do not need

tomdyson: > Ideally with the Play content removed or archived before deployment

removing is fine.

kaedroho: Will do, thanks!

PR #153, opened 2019-01-21
Cleanup

kaedroho:
- Remove body field from BlogPage and WorkPage (and rename streamfield to body)
- Remove "Play" logic
- Deleted TShirtPage model
- Deleted MarketingLandingPage model (and marketing logic)
- Move BlogPage intro into promote tab, rename it to "listing summary" and remove rich text features
- Remove StandardPage.middle_break and WorkPage.intro (neither were being used)
- Various other cleanups and renames


zerolab:
Worth adding this depends on #151

PR #163, opened 2019-03-19
[66] Add "strapline visible" checkbox

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/66


tbx/people/models.py
```diff
@@ -115,6 +115,7 @@ class CulturePageLink(Orderable):
 
 class CulturePage(Page):
     strapline = models.TextField()
+    strapline_visible = models.BooleanField()
```
kaedroho: Could we add help text mentioning that the strapline will still be in the frontend for accessibility reasons. Otherwise someone will end up removing it by mistake

tm-kn: Updated.

PR #164, opened 2019-03-19
[122] Allow adding more contact reasons list snippets

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/122


tbx/people/models.py
```diff
@@ -242,3 +242,12 @@ def __str__(self):
         FieldPanel('is_default', widget=forms.CheckboxInput),
         InlinePanel('reasons', label='Reasons', max_num=3)
     ]
+
+    def clean(self):
+        if self.is_default and \
+                ContactReasonsList.objects.filter(is_default=True).exists():
```
kaedroho: I think we need to exclude self from this query

tm-kn: Updated.

PR #166, opened 2019-03-21
coral theme implemented

NathHorrigan:


tbx/services/models.py
```diff
@@ -105,7 +105,12 @@ class ServicePageProcess(Orderable):
 
 class ServicePage(HeadlessPreviewMixin, Page):
     service = models.OneToOneField('taxonomy.Service', on_delete=models.SET_NULL, null=True, blank=True, help_text="Link to this service in taxonomy")
-    is_darktheme = models.BooleanField(default=False)
+    theme = models.CharField(max_length=255, choices=(
+        ('light', 'Light'),
+        ('coral', 'Coral'),
+        ('dark', 'Dark'),
+        ('dark--transparent', 'Dark with transparent header'),
+    ), default='light', blank=True, null=True)
```
kaedroho: There's no need for ``blank=True`` and ``null=True`` if there's a default.

helenb: @kaedroho can you make this change and push please?

tbx/services/migrations/0025_auto_20190320_1848.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.1.5 on 2019-03-20 18:48
```
kaedroho: Please don't forget to rename these migrations

PR #173, opened 2019-04-02
Add 404 page model and schema

NathHorrigan:


tbx/core/models.py
```diff
@@ -579,6 +580,34 @@ def serve(self, request, *args, **kwargs):
...
+        MultiFieldPanel(
+            [
+                SnippetChooserPanel('contact'),
+                SnippetChooserPanel('contact_reasons'),
+            ],
+            heading="Contact",
+            classname="collapsible"
+        ),
+    ]
+
```
kaedroho: Could you implement ``can_create_at`` to prevent more than one of these being created?

Something like
```python
@classmethod
def can_create_at(cls, parent):
    return super().can_create_at(parent) and not cls.objects.exists()
```

zerolab: Why not use https://docs.wagtail.io/en/v2.4/reference/pages/model_reference.html#wagtail.core.models.Page.max_count ?

zerolab:
No longer needed

PR #174, opened 2019-04-03
Redirects schema

NathHorrigan:


tbx/graphql/schema.py
```diff
@@ -487,6 +488,13 @@ class Meta:
         interfaces = [PageInterface]
 
 
+class RedirectObjectType(graphene.ObjectType):
+    site = graphene.String()
```
kaedroho: Do we need the ``site`` field?

tbx/graphql/schema.py
```diff
@@ -487,6 +488,13 @@ class Meta:
...
+    old_path = graphene.String()
+    link = graphene.String()
```
kaedroho: What about redirects that link to a page?

tbx/graphql/schema.py
tm-kn: no

tbx/graphql/schema.py
tm-kn: yes!

tbx/graphql/schema.py
```diff
@@ -699,6 +724,9 @@ def resolve_contact_reasons(self, info):
         except ContactReasonsList.DoesNotExist:
             return None
 
+    def resolve_redirects(self, info):
+        return Redirect.objects.select_related('redirect_page')
```
zerolab: This is missing an `.all()`

tbx/settings/dev.py
```diff
@@ -18,6 +18,7 @@
 
 AUTH_PASSWORD_VALIDATORS = []
 
+PREVIEW_URL = 'http://localhost:8003/preview/'
```
zerolab: Can you add a comment with a link to https://github.com/torchbox/torchbox-frontend for context?

tbx/graphql/schema.py
tm-kn: Why would you need `all`? I don't understand. :question: :confused:

tbx/settings/dev.py
tm-kn: This is from Nathan's code. I do not actually understand what it does.

tbx/graphql/schema.py
zerolab: `Redirect.objects` is just the manager, whereas all the other resolvers in the schema have either `.all()` or `.live()` which return the queryset. As far as I understand it you want to return the latter, but maybe I am just misreading the django graphene docs

tbx/settings/dev.py
zerolab: This configure the headless preview URL to point to the FE instance.

tbx/graphql/schema.py
zerolab: PEBCAK!

PR #175, opened 2019-05-07
[189] Add related case studies to a case study

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/189

Add related case studies that have the same tags to `caseStudies` GraphQL field.

FE that it enables to work - torchbox/torchbox-frontend#89


zerolab:
It does the job. :shipit:

PR #176, opened 2019-05-07
[202] Serialize rich text page links

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/202

Pass slug, service slug and page type so that the links on the Gatsby front-end can be created.

It is required by front-end - torchbox/torchbox-frontend#91

### DIFF
--- a/tbx/graphql/streamfield.py
+++ b/tbx/graphql/streamfield.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 
+from bs4 import BeautifulSoup
 from wagtail.core import blocks
 from wagtail.core.models import Page
 from wagtail.core.rich_text import RichText
@@ -7,8 +8,6 @@ from wagtail.embeds.blocks import EmbedBlock
 from wagtail.embeds.embeds import get_embed
 from wagtail.embeds.exceptions import EmbedException
 from wagtail.images.blocks import ImageChooserBlock
-
-from bs4 import BeautifulSoup
 
 
 class StreamFieldSerialiser:
--- a/tbx/graphql/streamfield.py
+++ b/tbx/graphql/streamfield.py
@@ -1,6 +1,8 @@
 from django.conf import settings
 
 from wagtail.core import blocks
+from wagtail.core.models import Page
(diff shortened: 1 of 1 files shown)


tm-kn:
Question for later: How do I use document.createElement with GatsbyJS? :confused:

tbx/graphql/streamfield.py
```diff
@@ -66,3 +68,38 @@ def serialise_block(self, block, value):
...
+                            'data-page-slug': page.slug,
+                            'data-page-service-slug': getattr(
+                                getattr(page, 'service', None), 'slug', None
+                            )
+                        }
+                    )
+                    new_tag.append(*anchor.contents)
+                    anchor.replace_with(new_tag)
+                except Page.DoesNotExist:
+                    new_tag = soup.new_tag('a')
```
zerolab: Should this not remove the link altogether?

tbx/graphql/streamfield.py
```diff
@@ -66,3 +68,38 @@ def serialise_block(self, block, value):
             return self.serialise_list_block(block, value)
         elif isinstance(block, blocks.StreamBlock):
             return self.serialise_stream_block(block, value)
+
+    def serialize_rich_text(self, source):
+        try:
+            from bs4 import BeautifulSoup
+        except ImportError:
+            return str(RichText(source))
```
zerolab: worth commenting why using `RichText()` here where previously it was just returning the source (well, text)

tbx/graphql/streamfield.py
tm-kn: I think that's default Wagtail behaviour. If the link is in the paragraph, the sentence may look stupid.

zerolab: fair enough

tm-kn:
@zerolab https://www.python.org/dev/peps/pep-0008/#maximum-line-length Following PEP8 but can change if you don't like it.

zerolab:
Won't nitpick more. LGTM

PR #178, opened 2019-05-09
[132] Strip out front end tooling

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/132

- Also set the home view to be 404.
- Wagtail views only accessible to the logged-in users.
- robots.txt to ignore the wagtail views as well.


tbx/urls.py
```diff
@@ -52,6 +54,7 @@
     path('review', include(wagtailreview_urls)),
```
zerolab: Do reviews still work with the headless/FE setup?

tbx/urls.py
```diff
@@ -71,8 +74,7 @@
...
 
 # Join private and public URLs.
-urlpatterns = private_urlpatterns + urlpatterns + [
-    # Add Wagtail URLs at the end.
-    # Wagtail cache-control is set on the page models's serve methods.
-    path(r'', include(wagtail_urls)),
-]
+urlpatterns = private_urlpatterns + urlpatterns + decorate_urlpatterns([
+    # Make wagtail paths required by login.
+    path('wagtail/', include(wagtail_urls))
```
zerolab: Shouldn't this change be documented/highlighted in the README?

Dockerfile
```diff
@@ -14,34 +14,15 @@ EXPOSE 8000
...
 RUN pip install -r requirements.txt
 
-# Copy front-end code
-WORKDIR /app/tbx/core/static_src
-COPY tbx/core/static_src .
-
-# Compile static files
-RUN npm run build:prod
-
-WORKDIR /app
```
zerolab: What about node in the VM? https://github.com/torchbox/wagtail-torchbox/blob/405e1a02480ab21b7da65ce26c8e20bdb4a3fcd4/vagrant/provision.sh#L67

tbx/urls.py
zerolab: btw, this is more of an observation, rather than something to do

tbx/urls.py
tm-kn: Updated. Better comment used as well.

tbx/urls.py
tm-kn: It should since it uses /admin/ URL. I'll test on staging once you clear the review.

Dockerfile
tm-kn: Cheers, sorted.

zerolab:
Do we need to leave the templates at all?

tm-kn:
> Do we need to leave the templates at all?

Let's leave them for now. Otherwise Wagtail "live" links will lead to HTTP 500 and not sure how that affects preview too. May be a good one for later.

tbx/urls.py
tm-kn: Previews work! :tada:

PR #179, opened 2019-05-09
[167] Update fabfile to pull only original images

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/167

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -11,6 +11,7 @@ STAGING_APP_INSTANCE = 'torchbox-staging'
 
 LOCAL_MEDIA_FOLDER = '/vagrant/media'
 LOCAL_DATABASE_NAME = 'torchbox'
+LOCAL_IMAGES_FOLDER = '/vagrant/media/original_images'
 
 
 ############
@@ -21,6 +22,11 @@ LOCAL_DATABASE_NAME = 'torchbox'
 @task
 def pull_production_media(c):
     pull_media_from_s3_heroku(c, PRODUCTION_APP_INSTANCE)
+
+
+@task
+def pull_production_images(c):
+    pull_images_from_s3_heroku(c, PRODUCTION_APP_INSTANCE)
 
 
 @task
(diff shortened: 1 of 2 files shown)


fabfile.py
```diff
@@ -111,6 +122,14 @@ def clean_local_database(c, local_database_name=LOCAL_DATABASE_NAME):
     create_local_database(c)
 
 
+def delete_local_renditions(local_database_name=LOCAL_DATABASE_NAME):
+    local(
+        'sudo -u postgres psql  -d {database_name} -c "DELETE FROM torchbox_torchboxrendition;"'.format(
```
helenb: I normally add a try / catch around this block, although if you definitely have the correct table name it probably isn't necessary.

helenb:
Looks good - have you tested it works on a local build?

fabfile.py
zerolab: Why not just `psql -Upostgres`?

zerolab:
One possible tweak, otherwise LGTM!

fabfile.py
tm-kn: This is from wagtail kit. You do not need to add try / catch block, you in fact should not. Fabric has API to deal with errors.

> psql -Upostgres

This won't work because the current setup uses peer authentication.

tm-kn:
@helenb I need someone else to test, it seems that generating renditions when Gatsby sends a lot of requests, results in a lot of unexpected errors.

helenb:
I'm afraid I don't have time to test today - does @thisoldbear have time to try it out? If it turns out this doesn't work with gatsby then we may have to forget it.

thisoldbear:
@helenb I've checked out this branch locally and am getting the following errors on the front end site, don't know if you also saw these @tm-kn?

Screenshot 2019-05-13 at 15 13 55

zerolab:
@tm-kn this looks to me quite deprecated. Shall we close it?

zerolab:
Note: this is already handled in the updated fabfile:

https://github.com/torchbox/wagtail-torchbox/blob/master/fabfile.py#L28
https://github.com/torchbox/wagtail-torchbox/blob/master/fabfile.py#L384

PR #180, opened 2019-05-09
[214] Add ID to jobs GraphQL schema

tm-kn:
https://projects.torchbox.com/projects/tbxcom-redesign/tickets/214


zerolab:
Worth adding IDs for all of the other models exposed to GraphQL?
:shipit:

PR #181, opened 2019-05-13
Update README.md

thisoldbear:
See thread in `#tech`

https://torchbox.slack.com/archives/C03R0FNK7/p1557761995251500

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ GraphQL: https://cms.torchbox.staging.torchbox.com/graphql/
 
 The automatic deployment should happen after merging to `staging`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,6 +81,8 @@ GraphQL: https://cms.torchbox.staging.torchbox.com/graphql/
 
 The automatic deployment should happen after merging to `staging`.
 
+Pushing changes to the `master` branch will automatically trigger a deployment to the production site.
```
zerolab: Why not merge this and "The automatic deployment should happen after merging to `staging`." into something like:

```markdown

### Deployments

Merges to `master` and `staging` will automatically trigger a deployment to the production and staging sites, respectively.
```

PR #203, opened 2020-08-07
Change JobsIndexPage to be fed via PeopleHR RSS instead of Wagtail

NathHorrigan:
This PR changes the `jobs` resolver function to use the PeopleHR RSS feed instead of Wagtail admin managed models.

When changes are made on PeopleHR a new Netlify build will need to be triggered via Admin > Netlify Deployments > New

FE: https://github.com/torchbox/torchbox-frontend/pull/123


tbx/graphql/schema.py
```diff
@@ -475,7 +480,34 @@ class JobsIndexPageObjectType(graphene.ObjectType):
     strapline = graphene.String()
 
     def resolve_jobs(self, info):
-        return self.jobs.all()
+        # Check xml feed is set
+        if self.jobs_xml_feed == "":
+            return []
+
+        # Get fresh data from People HR
+        res = requests.get(self.jobs_xml_feed)
```
kaedroho, line 488: We should add a timeout on this call, also we should make it handle network errors instead of crashing (maybe just return an empty list of jobs?). Make sure errors are still being reported to sentry if you do that though.

NathHorrigan: Good idea, Thanks

NathHorrigan:
Does that look like what you had in mind @kaedroho ??

tbx/graphql/schema.py
```diff
@@ -475,7 +481,42 @@ class JobsIndexPageObjectType(graphene.ObjectType):
...
+                    "url": job.find("link").getText(),
+                    "level": job.find("department").getText(),
+                    "location": html.unescape(
+                        job.find("city").getText()
+                    )
+                })
+
+            return jobs
+
+        except Exception as err:
```
kaedroho, line 514: Ideally, I think we would catch just network errors:

```python
from requests.exceptions import RequestException

try:
    ...
except RequestException as err:

```

kaedroho: But a more general note on catching all exceptions, don't forget to reraise `KeyboardInterrupt` and `SystemExit`:

```python
try:
    ...
except (KeyboardInterrupt, SystemExit):
    reraise
except Exception as err:
    ...
```

NathHorrigan: I'll switch to RequestException

PR #211, opened 2020-10-20
Back to django templates

zerolab:
* Strips out headlesspreview, netlify, graphql dependencies
* Adds django-pattern-library and WK default templates
* includes the WK navigation app
* Bump Django/Wagtail/Pillow/wbaip versions to include security releases


zerolab:
cc @nicklee

nicklee:
All works for me, thanks @zerolab

nicklee:
@zerolab do we need to install black on here and run it to format the BE files correctly so they pass CI?

zerolab:
I suggest we add black once WIPs are done.
We have flake8/isort. Linting fails here because of 
`tbx/core/templatetags/util_tags.py:7:1: E302 expected 2 blank lines, found 1` (meaning it needs another blank line above line 7)

PR #218, opened 2020-10-27
Redesign 2020: Expose Homepage fields

SharmaineLim:
No Codebase ticket yet, I think the homepage is being [redesigned](https://projects.torchbox.com/projects/tbxcom/tickets/276) before work is done on it.

This PR just exposes the current homepage fields and makes sure the model looks for the new HTML file than an HTML file under `templates/home/homepage.html`.

### DIFF
--- a/tbx/core/models.py
+++ b/tbx/core/models.py
@@ -227,6 +227,8 @@ class HomePageClient(Orderable, RelatedLink):
 
 
 class HomePage(Page):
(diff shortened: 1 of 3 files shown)


zerolab:
@chris-lawton is this still relevant given the new homepage work that went live last week?

chris-lawton:
Nope, it can be closed now as the all the homepage work was completed in https://github.com/torchbox/wagtail-torchbox/pull/331

PR #220, opened 2020-10-28
#262 - BlogPage: Link BE to FE (redesign-2020)

SharmaineLim:
[Codebase ticket #262](https://projects.torchbox.com/projects/tbxcom/tickets/262)

The focus of this PR is updating the FE from #213 to use existing model fields.


tbx/project_styleguide/templates/patterns/atoms/see-more/see-more.html
```diff
@@ -1,3 +1,6 @@
-<div class="see-more">
-    <a class="see-more__link" href="{{ item.href }}">{{ item.text }}</a>
-</div>
+{% load wagtailcore_tags %}
```
patrickcuagan, line 1: Based on the template below, I don't think this is needed.

tbx/project_styleguide/templates/patterns/molecules/blog-item/blog-item.html
```diff
@@ -1,25 +1,32 @@
...
+                    <span class="blog-item__author-name">{{ author.name }}</span>
+                {% if item.date %}
+                    <span class="blog-item__author-details">
+                {% endif %}
+                        <span class="blog-item__author-role">{{ author.role }}</span>
                 {% if item.date %}
-                    <span class="blog-item__author-date">{{ item.date }}</span>
+                        <span class="blog-item__author-date">{{ item.date }}</span>
+                    </span>
                 {% endif %}
```
patrickcuagan, line 27: The indentation here is very confusing... (though nothing can be done about it)

patrickcuagan (approved):
BE looks good to me! Just one comment about the FE but it's not major.

tbx/project_styleguide/templates/patterns/molecules/blog-item/blog-item.html
SharmaineLim: :joy: I know what you mean

tbx/project_styleguide/templates/patterns/atoms/see-more/see-more.html
SharmaineLim: Forgot to remove this from when I used `pageurl`. Thanks!

tbx/project_styleguide/templates/patterns/molecules/blog-item/blog-item.html
zerolab: the bigger question is why are we nesting these spans? are there more semantic tags that can be used?

SharmaineLim: I don't think there's a more semantic tag that can be used. `<div>` is an option but it also has no semantic meaning other than being a block by default than inline.

zerolab: I suppose not.. just feels odd, but alas

tbx/blog/models.py
```diff
@@ -156,6 +156,17 @@ def set_body_word_count(self):
         body_words = body_text.translate(body_text.maketrans(dict.fromkeys(remove_chars))).split()
         self.body_word_count = len(body_words)
 
+    @property
+    def related_blogs(self):
```
zerolab, line 160: nitpick: a blog is a collection of blog posts or blog entries. This should really be `related_posts`. (this is a hill I will quite likely die on 😂)

SharmaineLim: I'll integrate updating this to the PR I'm working on. 🙂

zerolab: 🌟

PR #231, opened 2020-11-05
Remove CORS headers

SharmaineLim:
To fix CORS errors on staging:

```
  File "/app/tbx/settings/base.py", line 7, in <module>

    from corsheaders.defaults import default_headers

ModuleNotFoundError: No module named 'corsheaders'

The command '/bin/sh -c SECRET_KEY=none python manage.py collectstatic --noinput --clear' returned a non-zero code: 1
```

We don't use corsheaders in wagtail-kit so it should be fine to remove.

Also, from Tom U.: "OK just checked, yes you can safely remove line 7, 76 and 464-469 from https://github.com/torchbox/wagtail-torchbox/blob/redesign-2020/tbx/settings/base.py"

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,6 @@ django-storages==1.6.6
 -e git://github.com/wagtail/wagtail-review.git@8ec5ec188ee75a075bd83c3a7804aeac9abf71be#egg=wagtail-review
 django-phonenumber-field==2.1.0
 phonenumbers==8.10.3
-django-cors-headers==3.0.2
 lxml==4.5.2
 django-pattern-library==0.2.8
 wagtail-accessibility==0.2.1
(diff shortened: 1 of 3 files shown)


zerolab:
Please remove the package from https://github.com/torchbox/wagtail-torchbox/blob/05b028dd2e73940c159a2c7f899732459a9f1cb0/requirements.txt#L16 too. Otherwise :shipit:

SharmaineLim:
> Please remove the package from
> 
> https://github.com/torchbox/wagtail-torchbox/blob/05b028dd2e73940c159a2c7f899732459a9f1cb0/requirements.txt#L16
> 
> too. Otherwise :shipit:

Missed that. Thanks!

PR #237, opened 2020-11-13
#264 - JobIndex Page: Link BE to FE (redesign-2020)

SharmaineLim:
[Codebase ticket #264](https://projects.torchbox.com/projects/tbxcom/tickets/264)

The focus of this ticket is to update the new job listing page in the pattern library to use the correct variables, and also to make sure the BE serves the correct HTML file.

In addition, I've added some code to pull the People HR feed.


tbx/core/models.py
```diff
@@ -425,6 +428,22 @@ class JobIndexPage(Page):
...
+            feed_success = True
+        except Exception as e:
+            jobs = None
+            feed_success = False
+            raise e
+
+        return render(
+            request,
+            self.template,
+            {"page": self, "jobs": jobs, "feed_success": feed_success},
```
NathHorrigan, line 444: Suggestion: Instead of `feed_success ` and populating `jobs` as `None`, I'd populate jobs as `[]` and then you can return something like:

```
return render(
             ...
             {"page": self, "jobs": jobs, "feed_success": len(jobs) > 0},
             ...
)
```

tbx/project_styleguide/templates/patterns/pages/job/job_listing.html
```diff
@@ -5,22 +5,31 @@
...
+                    {% for job in jobs %}
+                        {% include "patterns/molecules/job-item/job-item.html" with item=job %}
+                    {% endfor %}
+                </div>
+            {% else %}
+                <div>
+                    <p>
+                        An error occurred while getting this information, please try
+                        again.
+                    </p>
```
NathHorrigan, line 29: Is this the final text? I feel like "There is no jobs available right now, We update this page regularly so please come back" would be more appropriate. This is more of a question for design/DM but worth bringing up before merging.

tbx/core/api.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class PeopleHRFeed(object):
+    """
+    Pulls from PeopleHR RSS feed.
+    """
+
+    def get_jobs(self, *args, url=None, **kwargs):
+        if not url:
+            return None
```
NathHorrigan, line 13: If you switch to my suggestion then this will need to become `[]`

NathHorrigan (approved):
Thanks Amanda!

Just a few suggestions but if you short on time then feel free to merge.

BrianXu20:
Thank you @NathHorrigan,

All the suggestions are being implemented. 

This MR now is merged

PR #244, opened 2020-11-18
Feature/296 Base Template: add favicon

BrianXu20:
[Ticket #296](https://projects.torchbox.com/projects/tbxcom/tickets/296)
Added favicon icon to the project


zerolab:
in WK we use https://realfavicongenerator.net/ which outputs several formats. Might be worth doing that for this too

PR #252, opened 2021-01-22
#300 allow change default contact

BrianXu20:
Ticket:
https://projects.torchbox.com/projects/tbxcom/tickets/300

Added WagtailAdminPageForm to validate Contact snippet


tbx/people/forms.py
```diff
@@ -0,0 +1,28 @@
...
+        if cleaned_data["default_contact"]:
+            # If the default contact existed
+            if default_contact_enabled_count == 1:
+                contact = self.save(commit=False)
+                # Only save if user is modifying the default contact
+                if not models.Contact.objects.filter(
+                    default_contact=True, pk=contact.pk
+                ).exists():
+                    self.add_error(
+                        "default_contact", "There is an existed default contact."
```
alxbridge, line 25: **nitpick** (typo):

```suggestion
                        "default_contact", "There is an existing default contact."
```

alxbridge:
@BrianXu20 Your change all looks correct for fixing the current broken behaviour

**suggestion:** The usability of the behaviour could be improved beyond preventing a second default contact from being saved. Specific thoughts:

1. Might it be preferable to assume that the editor knows what they're doing, and just remove the 'default' setting from the existing contact before adding it to the current one?
2. Otherwise, it would be helpful to tell the editor in the error message who the default contact is, so that they're able to go in and remove the setting, without having to look through the list of users to find who the default is.

PR #259, opened 2021-02-01
In page nav

chris-lawton:
- Make in-page-nav sticky on scroll
- Highlight current section in nav when scrolling the page
- Move the process menu item higher up to match the order in which they appear in the DOM

### DIFF
--- a/tbx/static_src/sass/base/_base.scss
+++ b/tbx/static_src/sass/base/_base.scss
@@ -7,7 +7,7 @@
 
 html {
     &.template__service-page {
-        scroll-padding-top: 70px;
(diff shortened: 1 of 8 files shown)


tbx/static_src/sass/base/_base.scss
```diff
@@ -5,6 +5,12 @@
     box-sizing: border-box;
 }
 
+html {
+    &.template__service-page {
```
chris-lawton, line 9: ideally these would be refactored to use modifier BEM syntax `--` rather than the double underscore, but out of scope for this PR

PR #279, opened 2021-02-16
Bugfix/#310 work page sorting

BrianXu20:
- Display published date of work page
- Sort the work page by published date
ticket:https://projects.torchbox.com/projects/tbxcom/tickets/310


tbx/work/models.py
```diff
@@ -178,7 +180,12 @@ def get_popular_tags(self):
     @property
     def works(self):
         # Get list of work pages that are descendants of this page
-        return WorkPage.objects.descendant_of(self).live()
+        work_pages = WorkPage.objects.descendant_of(self).live()
+
+        # Order by most recent date first
+        work_pages = work_pages.order_by("-date", "pk")
```
alxbridge, line 186: **nitpick:** I think we want this to use `-pk` instead, so that if `date` isn't set then the items will still appear in the reverse order of when they were added.

tbx/work/models.py
```diff
@@ -67,6 +67,7 @@ class WorkPage(Page):
 
     parent_page_types = ["WorkIndexPage"]
 
+    date = models.DateField("Post date", null=True)
```
alxbridge, line 70: **question:** Is it correct that we should allow null post dates, rather than adding one for existing items in migration?

tbx/work/models.py
alxbridge: **suggestion (non-blocking)** If this is the ordering that should always be used for these items, it might be worth defining it in the model's Meta class with an [ordering](https://docs.djangoproject.com/en/3.1/ref/models/options/#ordering) option instead. But if other orderings might apply - or if this is the only place these objects are likely to be output - then doing this would probably be overkill.

alxbridge:
@BrianXu20 Just a couple of minor comments/suggestions, otherwise this all looks good.

tbx/work/models.py
BrianXu20: It keeps showing non-nullable field error. I think I have to try it again with database clean up

tbx/work/models.py
BrianXu20: As much as I want to use Meta class way but it doesn't work and I don't know why.
I will just use order_by instead

chris-lawton:
@alxbridge @BrianXu20 i think this has caused deployments to fail on staging and production. The logs are showing `django.db.utils.IntegrityError: column "date" contains null values`. Any ideas?

PR #286, opened 2021-02-17
Bugfix/#310 work page sorting

BrianXu20:
Ticket :
https://projects.torchbox.com/projects/tbxcom/tickets/310


tbx/work/models.py
```diff
@@ -178,7 +180,12 @@ def get_popular_tags(self):
     @property
     def works(self):
         # Get list of work pages that are descendants of this page
-        return WorkPage.objects.descendant_of(self).live()
+        work_pages = WorkPage.objects.descendant_of(self).live()
+
+        # Order by most recent date first
+        work_pages = work_pages.order_by("-date")
```
zerolab, line 186: suggestion: given we get empty dates this sort won't do any good. why not add a data migration that would populate the date field with either first_published_at, or last_published_at ?

BrianXu20: Thank you for the suggestion.
This actually makes a lot more sense.
I will refer your suggestion to the delivery manager.
Thank you.

BrianXu20: I have added data migration for this.
Thank you @zerolab

PR #288, opened 2021-02-18
[284] Font refactor

chris-lawton:
https://projects.torchbox.com/projects/tbxcom/tickets/284

Refactors almost all fonts to use the font-size mixin. There are a few instances where an explicit font size it still set on the large breakpoint as they didn't fit into, or were near any of the values in the large array.

On staging.


tbx/static_src/sass/components/_blog-item.scss
```diff
@@ -16,45 +16,14 @@ $blog-author: var(--testimonial-name);
         }
     }
 
-    &--featured {
```
nicklee, line 19: nice, guessing this wasn't used anywhere?

tbx/static_src/sass/components/_help-block.scss
```diff
@@ -50,16 +50,20 @@
...
 
         @include media-query(medium) {
+            @include font-size(xl);
             padding-left: 17vw;
-            font-size: 36px;
             line-height: 50px;
         }
 
+        @include media-query(large) {
+            @include font-size(l);
```
nicklee, line 64: when this is used within a breakpoint does it not confuse it? I thought we'd need something like `font-size: map-get($large-font-sizes, l);` otherwise it would output three media queries for each media query?

nicklee:
Nice one dude, great work, looks like you managed to strip some un-needed code too, bonus! couple of comments above

tbx/static_src/sass/components/_blog-item.scss
chris-lawton: couldn't see the class in the markup and it looks like the styles had been moved to a parent class - https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/static_src/sass/components/_blog-item.scss#L104

tbx/static_src/sass/components/_help-block.scss
chris-lawton: yup, good shout!

PR #290, opened 2021-02-26
Raw HTML block

chris-lawton:
Add styles for the Raw HTML block streamfield, was previously unstyled 

On staging: https://torchbox.staging.torchbox.com/work/brand-launch-for-breast-cancer-now/


helenb:
Changes LGTM @chris-lawton . Question: is this block mainly used for things like code snippets or is it more for things like embeds or custom html and js? Just wondering if those styles are appropriate for all use-cases.

chris-lawton:
Thanks @helenb. AFAIK it's more for custom html and js. I'll check in with Dex but I'm not 100% sure how much it actually gets used.

helenb:
Cool - looks like a good default for now.

PR #291, opened 2021-03-11
[325] Careers intro and image

chris-lawton:
Ticket: https://projects.torchbox.com/projects/tbxcom/tickets/325

Adds:

- Careers page intro styles and container sizing tweaks
- Sunday Times Top 100 image

Screenshot 2021-03-11 at 10 23 36


tbx/project_styleguide/templates/patterns/pages/culture/culture_page.html
```diff
@@ -21,7 +20,7 @@
             </div>
         {% endif %}
 
-        {% include "patterns/organisms/card-listing/card-listing.html" with card_items=page.links.all %}
+        <img src="{% static 'images/sunday-times-100-best-companies.jpg' %}" alt="Employees at the annual Torchfest event" loading="lazy">
```
laratorchbox, line 23: Are static used only for images in the static_src folder?

tbx/static_src/sass/components/_culture.scss
```diff
@@ -2,14 +2,20 @@
     $root: &;
 
     &__container {
-        @include container();
```
laratorchbox, line 5: What does @include container do?

tbx/static_src/sass/components/_culture.scss
```diff
@@ -2,14 +2,20 @@
...
-
-        @include media-query(large) {
-            @include container(medium);
-        }
+        margin: 0 auto;
+        max-width: 68vh;
 
         &--top {
-            padding-top: $gutter;
+            padding: ($gutter * 2) $variable-gutter--small 0;
```
laratorchbox, line 9: $grid: 20px;
$gutter: $grid;

Why do you not just write:
($grid * 2)

tbx/static_src/sass/components/_culture.scss
```diff
@@ -22,11 +28,15 @@
     }
 
     &__rich-text {
-        @include font-size(l);
+        font-size: map-get($small-font-sizes, m);
```
laratorchbox, line 31: How does map-get work?

laratorchbox (approved):
Looks all good - I just have a couple of CSS questions which I have left as comments. Thanks

tbx/project_styleguide/templates/patterns/pages/culture/culture_page.html
zerolab: The `static` template tag is used for anything in the `STATIC_ROOT` directory ([docs]( https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#static), so can be JS, CSS, fonts etc

In our case, everything `static_src` gets processed in some way (e.g. scss converted to css) and eventually ends up in the static directiy courtesy of npm tasks defined in `package.json`

tbx/static_src/sass/components/_culture.scss
chris-lawton: In Sass there's the concept of mixins (https://sass-lang.com/documentation/at-rules/mixin) which _allow you to define styles that can be re-used throughout your stylesheet._

In this repo there's `_mixins.scss` which contains a mixin called container - https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/static_src/sass/abstracts/_mixins.scss#L78. This mixin outputs a margin based on the `$size` that you pass to it.

To use that (or any) mixin you use `@include mixin-name()` so in this instance it's `@include container()` which will output the following css:

```
.culture__container {
    position: relative;
    margin: 0 8.33vw;
}

```

tbx/static_src/sass/components/_culture.scss
chris-lawton: We could do that but tend to use the variables for slightly different purposes. 

`$grid` is used for anything to do with alignment on the page (designs will usually have some sort of grid system).
`$gutter` is used for spacing like padding and margin.

(Out of interest i've just searched the codebase and the `$grid` variable is only being once so we could probably get rid of it)

tbx/static_src/sass/components/_culture.scss
chris-lawton: In `_variables.scss` are three maps of font-sizes: large, medium and small - https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/static_src/sass/abstracts/_variables.scss#L49

To set font sizes i'd usually `@include font-size(m);` which would output the value of `m` for each of the three breakpoints that correspond to the three maps - small: 22px, medium: 24px and large: 38px.

In this instance i don't want to out 3 values for 3 breakpoints, i just want to set one font size so `map-get($small-font-sizes, m);` will get the value of `m` in the `$small-font-sizes` map - https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/static_src/sass/abstracts/_variables.scss#L81

chris-lawton:
Thanks for the review @laratorchbox! : )

PR #293, opened 2021-03-11
#320 readtime computation error under preview mode

BrianXu20:
Ticket: https://projects.torchbox.com/projects/tbxcom/tickets/320
Removed signals (Wagtail doesn't provide a signal for draft submission 
Used save() instead 
Added Mixin

### DIFF
--- a/tbx/blog/models.py
+++ b/tbx/blog/models.py
@@ -4,7 +4,6 @@ import string
 from django import forms
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db import models
-from django.dispatch import receiver
 from django.shortcuts import render
 from django.utils.decorators import method_decorator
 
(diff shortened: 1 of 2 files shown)


tbx/core/utils/models.py
```diff
@@ -50,3 +54,33 @@ class SocialMediaSettings(BaseSetting):
...
+        remove_chars = string.punctuation + "“”’"
+        body_words = body_text.translate(
+            body_text.maketrans(dict.fromkeys(remove_chars))
+        ).split()
+        self.body_word_count = len(body_words)
+
+    @property
+    def read_time(self):
+        if self.body_word_count is None:
+            self.save()
```
tbx-sysadmin, line 85: **issue** This leads to unexpected behaviour when previewing a page which hasn't yet been saved: the draft is saved and (from my local testing as an admin user at least) also published. A simpler way of handling this might be to return a placeholder string for these cases - e.g. so that the preview shows 'xx min read'.

alxbridge: (Sorry - I was signed in as the wrong user - but the above comment was from me...)

BrianXu20: Can't believe that this suggestion is so simple and yet this PR is complicated.
But good idea for this one 
Thanks

alxbridge:
Thanks @BrianXu20 - this all looks good to me now

PR #298, opened 2021-03-12
[331] Contact block

chris-lawton:
https://projects.torchbox.com/projects/tbxcom/tickets/331

This wraps the page-contact template in a contact block so it can be overridden and hidden on the culture page template


nicklee (approved):
Super!

PR #307, opened 2021-03-23
[332] Careers page nav update

chris-lawton:
https://projects.torchbox.com/projects/tbxcom/tickets/332

Swaps out the culture nav item for the new careers page. 

As we have a hardcoded nav the page will need to be created/updated in wagtail when this goes live.


nicklee:
ooo this is a tricky one, not sure I can approve it, too complex ;)

nicklee (approved):

PR #308, opened 2021-03-23
Careers page streamfields

chris-lawton:
Ensures that the current streamfields work with the new widths that the careers page has brought with it. Only required for streamfield headings, paragraph and intro. 

`$wrapper--small` wasn't being used so i've repurposed it


nicklee (approved):
Looks great!

PR #309, opened 2021-03-23
fix docker-compose and fabfile to enable pulling DB backups

richbrennan:


fabfile.py
```diff
@@ -437,11 +437,11 @@ def heroku_login(c):
     """
     Log into the Heroku app for accessing config vars, database backups etc.
     """
-    subprocess.call(["docker-compose", "exec", "utils", "heroku", "login", "-i"])
```
tomusher, line 440: Is the `-i` flag not required for the interactive login prompt?

richbrennan: It errors with the below. Removing the `-i` allows you to open a browser and authorise the CLI through that.

```heroku: Enter your login credentials
Email [email]: 
Password: ***************
 ›   Error: Your account has MFA enabled; API requests using basic authentication with email and password are not 
 ›   supported. Please generate an authorization token for API access. 
 ›
 ›   Error ID: mfa_required```

tomusher:
LGTM!

PR #310, opened 2021-03-24
[328] Standout list BE

richbrennan:
Implements https://projects.torchbox.com/projects/tbxcom/tickets/328

- Adds a new streamfield to the culture page for adding standout items (see attached screenshot)
- Renames culture page to careers page in wagtail admin
- Updates shared templates and associated yaml files
- Adds methods to the culture and service pages to format data for shared templates

Also adds PUDB to the build for delightful debugging.

Screenshot from 2021-03-25 10-10-18

PR #312, opened 2021-03-25
[329] Latest blogs BE - careers page

richbrennan:
Implements https://projects.torchbox.com/projects/tbxcom/tickets/329

Adds:

- Features blogs to careers (culture) page
- Methods to pass blogs to template
- Updates to templates and YAML files

~There's a missing migration to add later, to avoid a clash with https://github.com/torchbox/wagtail-torchbox/pull/310~


tbx/blog/models.py
```diff
@@ -212,6 +212,14 @@ def blog_index(self):
     def has_authors(self):
         return self.authors.exists()
 
+    @property
+    def first_author(self):
+        """Safely return the first author if one exists."""
+        try:
+            return self.authors.first().author
+        except Exception:
```
BrianXu20, line 220: A generic exception might mean a lot of things,
Could be ValidationError, FieldDoesNotExist, or other things.
It doesn't really mean that author does not exist.
I think the follow code make more sense

```
author = self.authors.first()
if author:
    return author.author
return None   
```

tbx/people/models.py
```diff
@@ -146,8 +156,38 @@ class CulturePage(Page):
...
+        """Format the featured blog posts for the template."""
+
+        return [
+            {
+                "title": x.value.title,
+                "url": x.value.url,
+                "author": x.value.first_author,
+                "date": x.value.date,
+            }
+            for x in self.featured_blog_posts
```
BrianXu20, line 179: should always try to avoid using xyz or abc variables.
They are not clear for its purpose, in this case, name it blog_post or something similar would be better

tbx/services/models.py
```diff
@@ -134,19 +133,27 @@ class BaseServicePage(Page):
...
-        return render(
-            request,
-            self.template,
+        return [
             {
-                "page": self,
-                "blog_index_page": blog_index_page,
-                "work_index_page": work_index_page,
-            },
+                "title": x.blog_post.title,
```
BrianXu20, line 141: Please refer to my previous comment

BrianXu20 (approved):
The MR will work, just some syntax could be improved 
If you fix them then I think it's good for merge

tbx/people/migrations/0022_auto_20210326_1404.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 2.2.17 on 2021-03-26 14:04
```
zerolab, line 1: nitpick: we should avoid autogenerated migration names

PR #314, opened 2021-04-01
[333] Culture naming

chris-lawton:
https://projects.torchbox.com/projects/tbxcom/tickets/333

Updates:
- mobile nav to include careers page
- culture page template and css class names to careers
- how the heading colours work on the careers page - any `<span>` of `.title-block__heading` uses the color-accent var so we can just use that rather than overwrite.

### DIFF
--- a/tbx/people/models.py
+++ b/tbx/people/models.py
@@ -102,8 +102,9 @@ class CulturePageLink(Orderable):
     ]
 
 
+# Was previously the culture page until it was re-purposed to be the careers page
 class CulturePage(Page):
-    template = "patterns/pages/culture/culture_page.html"
+    template = "patterns/pages/careers/careers_page.html"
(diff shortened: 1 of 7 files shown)


tbx/people/models.py
```diff
@@ -102,8 +102,9 @@ class CulturePageLink(Orderable):
     ]
 
 
+# Was previously the culture page until it was re-purposed to be the careers page
```
chris-lawton, line 105: Chatted with Rich and we agreed that the best thing to do re. naming would be to rename the template name and classes but leave the model names as they are as they're tied to the DB tables and could be a faff to update

nicklee:
replace all culture > careers, sorted ;) all looks good to me

nicklee (approved):

PR #315, opened 2021-04-01
Unregister all service workers

tomusher:
Some users still have the old Gatsby service worker registered which is causing various strange caching issues.

This forces all service workers to be unregistered when a users visits the site.

### DIFF
--- a/tbx/static_src/javascript/main.js
+++ b/tbx/static_src/javascript/main.js
@@ -10,10 +10,10 @@ import InPageNav from './components/in-page-nav';
 
 import '../sass/main.scss';
 
-navigator.serviceWorker.getRegistrations().then(function(registrations) {
-    for(let registration of registrations) {
+navigator.serviceWorker.getRegistrations().then(function (registrations) {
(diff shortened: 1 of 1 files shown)


tbx/static_src/javascript/main.js
```diff
@@ -10,6 +10,12 @@ import InPageNav from './components/in-page-nav';
 
 import '../sass/main.scss';
 
+navigator.serviceWorker.getRegistrations().then(function (registrations) {
```
zerolab, line 13: the code itself LGTM, I do wonder if it should be wrapped in a `if ('serviceWorker' in navigator) {}` ?

zerolab: this would only affect very old browsers, IE in particular. Don't have access to the browser breakdown, but some public services are stuck on good ole' IE :)

zerolab (approved):
:shipit:

tbx/static_src/javascript/main.js
tomusher: Thanks Dan!

PR #294, opened 2021-03-11
Integration/culture page

chris-lawton:
Sprint: https://projects.torchbox.com/projects/tbxcom/sprints/be763855-f86f-35df-adb3-bfcbf00a8291/kanban

Contains:
https://github.com/torchbox/wagtail-torchbox/pull/292 - https://projects.torchbox.com/projects/tbxcom/tickets/324
https://github.com/torchbox/wagtail-torchbox/pull/291 - https://projects.torchbox.com/projects/tbxcom/tickets/325
https://github.com/torchbox/wagtail-torchbox/pull/297 - https://projects.torchbox.com/projects/tbxcom/tickets/327
https://github.com/torchbox/wagtail-torchbox/pull/295 - https://projects.torchbox.com/projects/tbxcom/tickets/326
https://github.com/torchbox/wagtail-torchbox/pull/307 - https://projects.torchbox.com/projects/tbxcom/tickets/332
https://github.com/torchbox/wagtail-torchbox/pull/298 - https://projects.torchbox.com/projects/tbxcom/tickets/331
https://github.com/torchbox/wagtail-torchbox/pull/304 & https://github.com/torchbox/wagtail-torchbox/pull/312 - https://projects.torchbox.com/projects/tbxcom/tickets/329
https://github.com/torchbox/wagtail-torchbox/pull/305 & https://github.com/torchbox/wagtail-torchbox/pull/310 - https://projects.torchbox.com/projects/tbxcom/tickets/328
https://github.com/torchbox/wagtail-torchbox/pull/327 - https://projects.torchbox.com/projects/tbxcom/tickets/313

PR #319, opened 2021-04-13
[282] Values page BE

chris-lawton:
Adds the BE for:

- A values page
- Title, strap line, intro, values, standout items and featured blog posts

It borrows heavily from the work done in https://github.com/torchbox/wagtail-torchbox/pull/294 for the standout items and the featured blog posts.


tbx/people/models.py
```diff
@@ -136,6 +150,100 @@ class CulturePage(Page):
...
+        [("blog_post", blocks.PageChooserBlock(page_type="blog.BlogPage"))],
+        blank=True,
+        verbose_name="Blog posts",
+    )
+
+    content_panels = [
+        FieldPanel("title", classname="full title"),
+        FieldPanel("strapline", classname="full"),
+        FieldPanel("intro", classname="full"),
+        MultiFieldPanel([InlinePanel("values", label="Values")], heading="Values"),
```
SharmaineLim, line 172: InlinePanel does this weird thing, where inside a MultiFieldPanel, it doesn't have a field label.

But if you keep it on its own and add a heading, it has a field label.

```
    InlinePanel("values", heading="Values", label="Values"),
```

`label` will add text to your "Add" button. So if you don't add `label`, the button will just say "Add", but if you add `label="Values"`, the button will say "Add values".

SharmaineLim (approved):
Looking good! Just one super-minor thing.

tbx/people/models.py
chris-lawton: Ah that's good to know, thanks!

PR #320, opened 2021-04-13
Fix: Media and dump dir in containers

tbrlpld:
I was having issues with pulling staging images and data. 

It turned out that the directories mounted on the `utils` container belonged to the `root` user because they where placed in the filesystem root. 

I have adjusted the paths to mount these directories in the `/app` directory (`/app/media` and `/app/database_dumps`). Now these directories can be written by the `tbx` user in the `utils` container (which is the default user that is running the fab commands).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,7 +74,7 @@ services:
     tty: true
     volumes:
       - ./media:/app/media:delegated,rw
-      - ./database_dumps:/database_dumps:delegated,rw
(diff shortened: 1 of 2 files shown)


tm-kn (approved):

tbrlpld:
Thanks for the review @tm-kn

zerolab (approved):

tbrlpld:
Thanks for review @zerolab

PR #322, opened 2021-04-14
Fix: Command to delete image rendition in fabfile

tbrlpld:
Because the app is using a custom rendition model, the command needed to be adjusted from the command in Wagtail kit (which I had just recently copied over here).


zerolab (approved):

tbrlpld:
Thanks @zerolab for reviewing

PR #327, opened 2021-04-19
[313] Careers page feedback

chris-lawton:
https://projects.torchbox.com/projects/tbxcom/tickets/313

Incudes:
- Adding gradient to hero image
- Include the card links at the top and bottom of the page
- Swap the order of the title and subtitle of standout items
- Remove body streamfield, contact block and strapline visible fields


tbrlpld:
**Suggestion** I can mention lines that where not changes in the PR, but Flake8 is complaining about unused imports:

```
tbx/people/models.py:23:1: F401 'wagtail.snippets.edit_handlers.SnippetChooserPanel' imported but unused
tbx/people/models.py:27:1: F401 'tbx.core.blocks.StoryBlock' imported but unused
```

These imports can be remove.

tbx/static_src/sass/components/_hero.scss
```diff
@@ -112,5 +112,18 @@
                 min-height: 700px;
             }
         }
+
+        &--image-overlay {
```
tbrlpld, line 116: **Question** I am totally new to BEM. I though `--` is for varying status like "hover" etc? Do we also use it as a component separator?

tbrlpld (commented):
Only change necessary is removing the unused import to fix the CI. Apart from that looks good to me 👍

tbx/static_src/sass/components/_hero.scss
chris-lawton: At Torchbox we tend to just stick with using `:hover` and `:focus` for styles relating to those states.

The `--` in BEM represents a modifier, which signifies an adjustment in styles on a Block or Element. In this instance i'm adding a modifier on the `hero__full-screen` element, so the final class is `hero__full-screen--image-overlay`.
https://css-tricks.com/bem-101/ is a good intro : )

chris-lawton:
Thanks @tbrlpld!

tbx/static_src/sass/components/_hero.scss
tbrlpld: Thanks @chris-lawton. That makes sense.

PR #328, opened 2021-04-29
Feature: Upgrade Wagtail to 2.13rc2 (latest release candidate)

tbrlpld:
I have checked the [upgrade considerations](https://docs.wagtail.io/en/latest/releases/2.13.html#upgrade-considerations) but have found nothing that needs to be adjusted.

---

- [x] `StreamField` blank and required rules. Make sure `blank=True` is used for an optional `SteamField`
-> Nothing found
- [x] `StructBlock` overriding `get_form_context`
-> Nothing found
- [x] `FieldBlock` requiring JS functions on initialisation.
-> Nothing found
- [x] SVG icons on `register_setting` and `register_settings_menu_item` only needed when using custom icon fonts. 
-> Nothing found
- [x] `settings_panel` if not reusing the `Page.settings_panel` but commenting is desired, `CommentPanel` needs to be added to the list explicitly. 
-> nothing found

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 .DS_Store
 *.swp
 /venv/
+.env
+.python-version
 /static/
 static_compiled/
 /media/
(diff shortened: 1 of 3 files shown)


tomdyson (approved):

tbrlpld:
Thanks for having a look @tomdyson 👍

PR #334, opened 2021-05-21
[312] Events page BE

chris-lawton:
Adds the backend for the events based on https://projects.torchbox.com/projects/tbxcom/tickets/312#update-67103629

Screenshot 2021-05-21 at 16 35 33

TODO: 
- [x] Add the filters on the index page
- [x] Add the filters to the individual events
- [ ] Allow for multiple authors to be added to an event (unsure if this is required for MVP)


richbrennan:
I haven't allowed for multiple authors as the template is set to work with one, and comment above says unsure if this is required for MVP

tbx/events/models.py
```diff
@@ -0,0 +1,72 @@
...
+    def get_context(self, request):
+        context = super().get_context(request)
+        context.update(
+            events=self.get_events(request.GET.get("filter")),
+            related_services=Service.objects.all(),
+        )
+        return context
+
+
+class Event(ClusterableModel, Orderable):
```
kaedroho, line 46: I haven't seen a clusterable child model before, do you know if this has been done elsewhere?

tbx/project_styleguide/templates/patterns/pages/events/events_listing.html
```diff
@@ -0,0 +1,22 @@
...
+{% block content %}
+
+    {# Title and Filters #}
+    {% include "patterns/molecules/title-block/title-block.html" with item=page tags=related_services %}
+
+    {# Events #}
+    <div class="events">
+        <div class="events__container {% if not events %}events__container--empty{% endif %}">
+            {% for event in events %}
+                {% include "patterns/molecules/blog-item/blog-item.html" with item=event classes="blog-item--standout" %}
```
kaedroho, line 14: I think we should make a copy of this template for events, so this feature is decoupled from the blog. This should reduce any risk of a change being made to that template that could break the events page.

kaedroho (changes requested):
One question and a requested change, but otherwise looks good to me!

tbx/events/models.py
richbrennan: Nope, hadn't realised I'd done it, thanks for spotting!

richbrennan: It turns out I do need this, otherwise the `ParentalManyToManyField` doesn't work.

kaedroho: I have a feeling that this isn't supported in Wagtail. It probably looks fine in the editor but certian features, like copying, might not work

phildexter: just spoke to Karl. This would need a fix in wagtail core to get this many to many in an inline panel working perfectly.

At the moment we've agreed we can deploy this as is, and just know that if we copy the page we'll not have the same content on the copied page (can't see this being an issue here).

PR #339, opened 2021-07-07
Add new Process section CTA to BaseServicePage in process-block

thibaudcolas:
The intention with this is to allow for a more flexible call to action than what the “Contact” fields allow. We need this for an Accessibility Services SubServicePage.

Here is what this looks like in the CMS:

localhost_8000_admin_pages_5_edit_ (2)

And on the site’s frontend – the intention is for the text styles to match those of the blocks above, but the spacing should be more like the full-width text content further up the page.

localhost_8000_admin_pages_5_edit_preview_

---

## Testing

The UI for this can be tested in the pattern library at http://localhost:8000/pattern-library/pattern/patterns/organisms/process-block/process-block.html.

In the CMS, this should be tested on a SubServicePage. Here is the one we are expecting to use this on once available: https://torchbox.com/admin/pages/1460/edit/.

Test content we’re using, to copy-paste:

<details>

<summary>Test content</summary>

## Find out more…

Every digital platform or service is unique, so we’ll review your site against your objectives and provide bespoke recommendations suitable for your organisation. There is no ‘one size fits all’ when it comes to accessibility testing.

If you’d like to know how to improve the accessibility of your digital product or service, drop us your details with a link to the site(s) to test (if possible) and we’ll get back to you with our recommended approach.

Make your digital platform the best it can possibly be – email [test@example.com](mailto:test@example.com)

</details>

Tested only locally in macOS Chrome, as the styles are very simple.


alxbridge:
Hi @thibaudcolas - this all looks fine & works well for me locally

alxbridge (approved):

thibaudcolas:
Thank youuuuuuuuuuuuuuuuuuu

PR #341, opened 2021-07-15
Restoring password required template

nicklee:
This restores the password required template that was originally added then removed when we ditched most of WK templates


zerolab:
The change LGTM, but the page itself needs some love

<details>
<summary>📸 screenshot</summary>

Screenshot 2021-07-15 at 10 02 32
</details>

zerolab:
Happy to get this out as to fix the 500 error, but we need to show those styles some ❤️

nicklee:
Thanks Dan!

@phildexter sounds like we need to get a new ticket to design and then style this page up

PR #343, opened 2021-07-19
#353 Add min_num for authors on blog authors and case studies

stevedya:
Ensure at least 1 author is required on blog posts and case study posts


zerolab (approved):
LGTM

PR #345, opened 2021-07-20
[355] Add related case studies to profiles and blogs

stevedya:
- Add 2 work pages related by the same author to profile pages
- Add 2 work pages related by the same service to blog pages


helenb:
@stevedya  FE changes LGTM. I think a back-ender should probably double check the python changes (but all working on my local build)

tbx/blog/models.py
```diff
@@ -216,6 +217,19 @@ def related_blog_posts(self):
             .exclude(pk=self.pk)[:2]
         ]
 
+    @property
+    def related_works(self):
+        services = self.related_services.all()
+
+        # Get 2 work pages with the same service
```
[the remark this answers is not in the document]
stevedya: Perfect :) I have made that change now

tbx/people/models.py
```diff
@@ -82,6 +83,20 @@ def author_posts(self):
...
         ]
 
+    @property
+    def related_works(self):
+        # Get author
+        author_snippet = Author.objects.get(person_page__pk=self.pk)
+
+        # Get the latest 2 work pages by this author
+        works = (
+            WorkPage.objects.filter(authors__author=author_snippet)
```
zerolab, line 93: suggestion: 
```suggestion
            WorkPage.objects.filter(authors__author__person_page=self.pk)
```

should remove the need for `author_snippet = Author.objects.get(person_page__pk=self.pk)`

tbx/people/models.py
```diff
@@ -82,6 +83,20 @@ def author_posts(self):
             .order_by("-date")
         ]
 
+    @property
```
zerolab, line 86: may be worth making it a `cached_property` here and in blog

zerolab (approved):

tbx/blog/models.py
```diff
@@ -216,6 +218,19 @@ def related_blog_posts(self):
...
         ]
 
+    @cached_property
+    def related_works(self):
+        services = self.related_services.all()
+
+        # Get the latest 2 work pages with the same service
+        works = (
+            WorkPage.objects.filter(related_services__in=services)
+            .live()
```
zerolab, line 228: I've been thinking... we should probably also do `.public()` for all of these related pages in case any of them is restricted in any way.

chris-lawton: is this literally just a case of doing 
```
WorkPage.objects.filter(related_services__in=services)
            .live().public()
```
?

zerolab: Indeed

PR #348, opened 2021-07-23
[330] Instagram rework

chris-lawton:
Changes the current instagram implementation to be a manual process

Screenshot 2021-07-23 at 16 47 37

@tbrlpld Are you able to review this? Also i've removed some of the code you had in previously the pipelines are failing as the migrations are referencing bits that don't exist anymore (i think), any ideas on how to fix that?


tbx/people/blocks.py
```diff
@@ -42,60 +42,9 @@ def get_link(value):
...
-            return ""
-
-        return render_to_string(
-            template_name="patterns/atoms/instagram-post/instagram-post.html",
-            context={"embed": embed},
-        )
-
+class InstagramPostsBlock(blocks.StructBlock):
+    image = ImageChooserBlock()
+    link = blocks.URLBlock(required=False, help_text="Link to a specific post here or leave blank for it to link to https://www.instagram.com/torchboxltd/")
```
tbrlpld, line 47: **Suggestion** You can set the default value with `default="https://www.instagram.com/torchboxltd/"`. This way you can remove the logic from the template.

tbrlpld (commented):
@chris-lawton This is coming along nicely. Thanks for jumping in on this. 

I am not quite sure why the gallery block element with its template was deleted? Before, there was a single block that needed to be added to the page type and a template that would automatically render that block and could be used in the pattern library. Now, the block definition needs to be on the page type and the page template needs to contain more logic to render the gallery. 

Just wondering why this previously existing work was removed?

tbrlpld:
@chris-lawton Regarding the failing pipeline, I think there are just some imports left over that are now unused and `flake8` does not like that:

```
tbx/people/blocks.py:1:1: F401 'django.core.exceptions' imported but unused
12
tbx/people/blocks.py:2:1: F401 'django.template.loader.render_to_string' imported but unused
13
tbx/people/blocks.py:3:1: F401 'django.utils.functional.cached_property' imported but unused
14
tbx/people/blocks.py:6:1: F401 'wagtail.embeds.blocks as embed_blocks' imported but unused
15
tbx/people/blocks.py:7:1: F401 'wagtail.embeds.embeds' imported but unused
16
tbx/people/blocks.py:8:1: F401 'wagtail.embeds.exceptions as embed_exceptions' imported but unused
```

chris-lawton:
@tbrlpld 

> I am not quite sure why the gallery block element with its template was deleted? Before, there was a single block that needed to be added to the page type and a template that would automatically render that block and could be used in the pattern library. Now, the block definition needs to be on the page type and the page template needs to contain more logic to render the gallery.

> Just wondering why this previously existing work was removed?

This is most definitely down to my lack of knowledge of BE on my part (i've only recently started to pick it up) - so this a good one to learn from in the future - i'm all for reusability 🙂

In the interest of putting it live and not creating more work for ourselves i'm going to leave it as is but thanks for pointing that out.

I've fixed all the CI linting errors, however i'm now getting `AttributeError: module 'tbx.people.blocks' has no attribute 'InstagramEmbedBlock'` presumably because i've deleted a block that is still being referenced by a previous migration? Any ideas how you'd handle this?

tbrlpld:
> I've fixed all the CI linting errors, however i'm now getting `AttributeError: module 'tbx.people.blocks' has no attribute 'InstagramEmbedBlock'` presumably because i've deleted a block that is still being referenced by a previous migration? Any ideas how you'd handle this?

Hm, yea that's a tricky one. That is always a bit annoying when dealing with custom StreamField blocks. I guess you could remove the import/reference in the migrations manually, but I would get another opinion on this one.

PR #349, opened 2021-07-28
[359] Aligned image SF quality

chris-lawton:
https://projects.torchbox.com/projects/tbxcom/tickets/359

The aligned image streamfield is currently outputting images at `max-730x490` and is set to be `max-width: 50vw;` which means they are very pixelated when used on a blog page and viewed on a wide screen. 

This MR resolves that by increasing the size of the image when the alignment option is set to full width.

(click the gif below to open in a new tab and see the difference in image quality)

tbx-align-image-sf


nicklee (approved):
Very niceeee

PR #351, opened 2021-07-29
Remove careers page link from the mobile menu

chris-lawton:


chris-lawton:
Closing as the careers page is now live (we were going to remove the link if it wasn't going to be live by EOD)

PR #354, opened 2021-08-24
[362] Order job listing alphabetically

SimonDEvans:
Opened a blank ticket ([362](https://projects.torchbox.com/projects/tbxcom/tickets/362)) to track this.

We should probably order the list on the BE after pulling the data via the PeopleHR API, but this is ok for now I think.


nicklee:
nailed it, nice one

nicklee (approved):

PR #357, opened 2021-11-01
Fix type for blog posts

zerolab:
This is a hill I am ready to die on :)

A blog is a collection of blog posts or blog entries.


RealOrangeOne (approved):
I will die on this hill right alongside you!

PR #359, opened 2021-11-01
Homepage video image

chris-lawton:
Currently the hero image is shown on mobile and desktop even though it's not used on desktop - this means you see a flash of the image before the video loads. This MR hides the image by default and only shows it if the video fails to load.

### DIFF
--- a/tbx/static_src/javascript/components/shards-video.js
+++ b/tbx/static_src/javascript/components/shards-video.js
@@ -17,6 +17,7 @@ class ShardsVideo {
         if (window.matchMedia('(min-width: 1023px)').matches) {
             this.videoSource.src = this.videoSource.getAttribute('data-src');
             this.video.load();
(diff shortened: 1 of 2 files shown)


tbx/static_src/javascript/components/shards-video.js
```diff
@@ -28,6 +28,11 @@ class ShardsVideo {
                 this.video.pause();
             }
         });
+
+        // Show a fallback image if the video fails to load
+        this.videoSource.addEventListener('error', () => {
```
jhancock532, line 33: This is cool! There's a few other video events that we might want to listen for, such as [`suspend`](https://www.w3schools.com/tags/av_event_suspend.asp), [`stalled`](https://www.w3schools.com/tags/av_event_stalled.asp) and [`abort`](https://www.w3schools.com/tags/av_event_abort.asp). `error` seems like a good generic choice though.

PR #365, opened 2021-11-12
Custom images app: migrate images (part 2)

William-Blackie:
Builds on #364


tbx/images/migrations/0002_copy_custom_images.py
```diff
@@ -0,0 +1,33 @@
...
+    # Here we copy and initialise our new image models and their renditions
+    CustomImage = apps.get_model("images", "CustomImage")
+    Rendition = apps.get_model("images", "Rendition")
+
+    TorchboxImage = apps.get_model("torchbox", "TorchboxImage")
+    TorchboxRendition = apps.get_model("torchbox", "TorchboxRendition")
+
+    # Copy old images to new model
+    for tbx_image in TorchboxImage.objects.values():
+        custom_image = CustomImage.objects.create(**tbx_image)
```
gasman, line 14: This is going to include the primary key in the fields passed to `CustomImage.objects.create` - which is clearly a good thing, because if we weren't preserving image IDs we wouldn't be able to switch over the foreign keys with a lot of faff. However, I'm a little bit hazy on what effect that's going to have on the sequence that allocates new IDs - I suspect it might leave it at ID 1, and I'm not sure if Django has any mitigation in place to avoid new IDs colliding with existing ones.

Could you check what happens with the IDs of newly-uploaded images after this migration has run - does it allocate new IDs above the old ones, or start again at 1 but gracefully skip existing ones? If it's neither of the above, we'll probably have to have a line here explicitly setting the sequence to max-id-plus-one.

(Same for renditions, although in that case we can get away with the simpler fix of not preserving the original ID.)

gasman (commented):
Just one query, which might just be me being overly paranoid... other than that, all three PRs look good to me!

tbx/images/migrations/0002_copy_custom_images.py
William-Blackie: Hi @gasman, 

Thanks for reviewing this. I've done a little investigating and your worry is correct, the ID is allocated from 1 upon saving a new image; failing for obvious reasons. I'll investigate and see if I can come up with something more elegant.

PR #367, opened 2021-11-15
Use `get` to access `job_count` in context

tbrlpld:
I was getting "KeyError" on all pages when working locally. I assume
because the API access to retrieve the job count is not setup correctly.

I have now protected the key access with `get` so that the access fails
more gracefully.

### DIFF
--- a/tbx/navigation/templatetags/navigation_tags.py
+++ b/tbx/navigation/templatetags/navigation_tags.py
@@ -13,7 +13,7 @@ def primarynav(context, is_home):
         "primarynav": context["settings"]["navigation"][
             "NavigationSettings"
         ].primary_navigation,
-        "job_count": context["job_count"],
+        "job_count": context.get("job_count", 0),
         "request": request,
         "is_home": is_home,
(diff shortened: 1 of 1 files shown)


tbrlpld:
Never mind. This was caused by my `local.py`.

zerolab:
@tbrlpld imho this is a sensible change.
It may also be worth adding a note in the setup section if the issue with your `local.py` is something that others may have

tbrlpld:
@zerolab The `local.py` was excluding the context manager that adds the `jobs_count` to the context. I think last time I worked on the project the job count was in refactoring and the context manager led to errors. This seems to be now fixed and the context manager works locally too, without having to configure anything. 

I have also reopened the PR is you are happy to review 😅

zerolab (approved):

tbrlpld:
Thanks @zerolab

tbrlpld:
@zerolab I have deployed to staging and the site still works 😅 Not sure how else to test that before release to production.

PR #368, opened 2021-11-16
Make the request as part of the exception handling

RealOrangeOne:
This ensure that read timeouts are also captured, as these would be captured _before_ `raise_for_exception` is called.


tomusher:
Looks good!

PR #370, opened 2021-12-06
Improve handling of location fields from PeopleHR jobs feed

alxbridge:
For [Support ticket 305](https://projects.torchbox.com/projects/support-team/tickets/305)

https://torchbox.com/jobs/ is currently broken, as one of the vacancies coming from the PeopleHR feed doesn't contain a `city` attribute which the code assumes will be present.

I've amended the code so that it will use either/both of the `city` and `country` attributes if present, and an empty string if neither are supplied.

### DIFF
--- a/tbx/core/api.py
+++ b/tbx/core/api.py
@@ -23,12 +23,19 @@ class PeopleHRFeed(object):
 
         for node in xml_root.iter("item"):
             job = {}
-
             job["title"] = node.find("vacancyname").text
             job["description"] = node.find("vacancydescription").text
             job["department"] = node.find("department").text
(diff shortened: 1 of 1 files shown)


nealtodd:
That should fix the immediate error with he city not being present. Looking at the feed the same text that would have normally gone into city has been put in country.  Wouldn't expect both fields to get the same text in one job so this approach of merging them into one string looks fine - in practice it looks like it'll be one or the other.

Outside the scope of this support ticket it will be worth the get_jobs method having the same defensive coding against failure/timeout of the RSS feed request as has been put in the get_job_count method, but that can be raised as an issue for whoever is currently maintaining the site.

PR #295, opened 2021-03-11
[326]  Benefits List

laratorchbox:
Ticket: https://projects.torchbox.com/projects/tbxcom/tickets/326

Adds:
- list of benefits 
- refactoring of code in order to reuse list items across culture page and help block.html

Screenshot 2021-03-11 at 16 11 42

PR #374, opened 2021-12-15
[295] Update Primary Navigation

jhancock532:
Following [ticket #295](https://torchbox.codebasehq.com/projects/support-team/tickets/295) this updates the primary navigation to the new design. 

I haven't adopted the new font as I felt this was out of sync of the rest of the website design. It might be worth updating the font weight to bold for the in page navigation menu items, although the current website displays these with normal weighting.

I've added a new conditional to the primarynav component, `is_desktop`. This renders the new condensed nav for desktop devices, while rendering the mobile nav menu as the original list.

I'm not sure why the careers badge notification icon has a link to the jobs page; this seems like an obscure interaction. I've kept it in the dropdown menu, but removed the link for the More button. This prevents the user from getting the same link twice while they are tabbing through the menu.

The new design has a translucent background for the nav menu. I haven't implemented this, as against colored backgrounds as it doesn't look good. It may be worth theming this so the subnav is translucent only against white backgrounds?

I've installed a new library to handle the scroll blocking, [body-scroll-lock](https://www.npmjs.com/package/body-scroll-lock). It's imported into the new primarynav.js component.

I've tested using browserstack on Windows with Edge, Safari on Mac, Chrome on Mac.


nickmoreton:
@jhancock532 This all looks good to me. I tested it locally and it's working OK. Nice work!

**Question** I'm not sure sure if the scroll lock was a requirement. It works as expected and I can see why you added it as the figma looks like it should be a feature. Although I would double check with Ben on that. I actually think the effect seen in Figma is just the way the preview works rather than a design requirement and we don't seem to have that feature at the home page when opening the `More` menu

jhancock532:
@nickmoreton I've confirmed with Ben that this is how the site should work, though he'd like to check out the effect on staging first before we make it live. I would also like to double check for any accessibility issues on staging with Jane, I feel the addition of a clear "Close" button to the menu might make this more user friendly

tbx/static_src/javascript/components/primary-nav.js
```diff
@@ -0,0 +1,131 @@
...
+
+class PrimaryNav {
+    static selector() {
+        return '[data-primary-nav]';
+    }
+
+    constructor(node) {
+        this.node = node;
+        this.open = false;
+        this.button = node.querySelector('.nav-item__button');
```
janehughes, line 11: Can we use data attributes rather than classes for the selectors? Lets keep classes for styling

tbx/static_src/sass/components/_badge.scss
```diff
@@ -14,7 +14,7 @@
         background: var(--color--header-icon-color);
         position: absolute;
         transform: rotate(30deg);
-        left: -5px;
+        left: -3px;
```
janehughes, line 17: if you make the parent badge class `position: relative`, you could use `left: 50%` `transform: translate(-50%)` to center the number horizontally.

alternatively, the magic of flexbox may be more elegant :)

janehughes (commented):
This is looking great so far James thanks! Only a couple of comments, will need to find out the final verdict on the jobs count badge too before proceeding with this.

tbx/static_src/sass/components/_subnav.scss
```diff
@@ -46,8 +46,38 @@
...
     &__badge-link {
         position: relative;
-        left: 10px;
+        left: 14px;
         bottom: 10px;
         border-bottom: 0;
     }
+
+    // Don't style homepage nav with new rules, follow old design
+    .template__homepage & {
```
jhancock532, line 55: Using this styling convention `.class &` as it follows what Chris has done for badge.scss. Not sure if in future this would be better to have as a seperate class modifier e.g. `class--home` and `class--default` or similar

nickmoreton:
@jhancock532 

**Question** Was there other work here you wanted me to review? You did answer my questions about the overlay :) 

**Suggestion** There seems to be other unresolved comments from @janehughes that could still be outstanding...

jhancock532:
After discussion on Slack we decided to remove the additional notification badge - I have done so in my most recent commit, and my fix: commits above have resolved Jane's other comments.

nickmoreton:
@jhancock532 LGTM 👍  Go for staging.

jhancock532:
Added to staging, awaiting internal review

jhancock532:
Reviewed by Lisa, merged with production.

PR #381, opened 2022-02-15
Support-353: Wagtail 2.16

BrianXu20:


BrianXu20:
Just run the poetry update, nothing really changed here at this point except poetry.lock

PR #382, opened 2022-02-22
Add fallback content if PeopleHR job listing fields are forgotten

jhancock532:
This should prevent the `/jobs` page from crashing if information from the PeopleHR feed is forgotten. I've added generic placeholder content for each field, please double check the copy I've used. 

I think that peopleHR is set up so that the link should always be present, as this isn't a user editable field. If peopleHR crashes or the feed breaks, I've added a mailto link that directs to recruitment@torchbox.com. Alternatively we could add a link that redirects to a 404 page, but in this scenario, it's less likely someone will reach out to contact us about the issue.

### DIFF
--- a/tbx/core/api.py
+++ b/tbx/core/api.py
@@ -33,7 +33,9 @@ class PeopleHRFeed(object):
             try:
                 job["description"] = node.find("vacancydescription").text
             except AttributeError:
-                job["description"] = "This role is currently being considered and will be announced shortly."
+                job[
+                    "description"
+                ] = "This role is currently being considered and will be announced shortly."
 
             try:
                 job["link"] = node.find("link").text
--- a/tbx/core/api.py
+++ b/tbx/core/api.py
@@ -23,10 +23,27 @@ class PeopleHRFeed(object):
 
         for node in xml_root.iter("item"):
             job = {}
-            job["title"] = node.find("vacancyname").text
-            job["description"] = node.find("vacancydescription").text
-            job["department"] = node.find("department").text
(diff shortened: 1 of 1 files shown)


tbx/core/api.py
```diff
@@ -23,10 +23,29 @@ def get_jobs(self, *args, url=None, **kwargs):
...
+                job["title"] = node.find("vacancyname").text
+            except AttributeError:
+                job["title"] = "Role to be Announced"
+
+            try:
+                job["description"] = node.find("vacancydescription").text
+            except AttributeError:
+                job[
+                    "description"
+                ] = "This role is currently being considered and will be announced shortly."
```
zerolab, line 38: question: should we not skip a role if any of these are not filled in?

if I understood the latest issue correctly, it was because of a missing address field?

RealOrangeOne: I agree. If a role is "To be announced" and has little to not useful detail on it, there's very little use in actually publicising it.

Adding some logging as to _why_ a position isn't listed would help us diagnose the problem, if there was a role which wasn't appearing.

jhancock532: I've updated this code to skip adding the job to the page if any of the data is missing, with additional error logging that highlights which job post in PeopleHR is missing required fields. I haven't added field specific error logging, should I copy paste the try-catch statements to cover this? 

Resolving PeopleHR issues should just be a case of double checking the job listing fields, so I've held off on this for now.

jhancock532:
This can now be closed as the Careers site integration has gone live.

PR #396, opened 2022-04-22
Upgrade Wagtail to 3.0rc1

nickmoreton:
# This upgrades Wagtail from v2.16 to v3.0rc1

Ticket: https://projects.torchbox.com/projects/support-team/tickets/426

### Agreed with other team members

- wagtail-storages isn't compatible with Wagtail > 2 e.g. Wagtail 3.0rc1 I have removed the package as it's not used on this site.

## Upgrade considerations

**Changes to module paths**

- `wagtail updatemodulepaths` has been run

**Removal of special-purpose field panel types**

- Changes have been made

**BASE_URL setting renamed to WAGTAILADMIN_BASE_URL**

- Updated

**use_json_field argument added to StreamField**

- Updated

## Upgrade considerations - changes affecting Wagtail customisations

- No Wagtail customisations found.


pyproject.toml
```diff
@@ -22,7 +22,7 @@ django-csp = "~3.6"
 scout-apm = "~2.19"
 django-basic-auth-ip-whitelist = "~0.3.3"
 wagtail-accessibility = "~0.2"
-wagtail-storages = "~0.0.4"
+# wagtail-storages = "~0.0.4"
```
BrianXu20, line 25: I am not super familiar with `wagtail-storages`, will this affect our S3 setup, like making it unusable? 
This might make sense to remove it if we are not using S3?

BrianXu20 (commented):
I left a small concern.

pyproject.toml
zerolab: wagtail-storages asllows for ACL on docs in S3 when using a CDN
tl;dr it either uses the public S3 url for non-restricted docs, or the a signed S3 url for the private ones

zerolab: best to submit PR to https://github.com/torchbox/wagtail-storages with 3.0 restriction relaxed.

nickmoreton: Thanks Dan, I will submit a PR for that. Do you agree with my thoughts that it's not used on this site, I couldn't see any configuration for it?

BrianXu20: If we are not using it, then I am cool with this.

zerolab: Looks like it came with the move to poetry based on kit. remove and let's keep moving

PR #404, opened 2022-08-24
FE proposition page sections

helenb:
See https://projects.torchbox.com/projects/tbxcom/tickets/392

PR #405, opened 2022-08-25
Integration branch for new proposition landing page

helenb:
Includes the following PRS:

https://github.com/torchbox/wagtail-torchbox/pull/403
https://github.com/torchbox/wagtail-torchbox/pull/404
https://github.com/torchbox/wagtail-torchbox/pull/407
https://github.com/torchbox/wagtail-torchbox/pull/415


tbx/services/models.py
```diff
@@ -320,13 +321,16 @@ class ServicePageProcess(Orderable, BaseServicePageProcess):
 class SubServicePage(BaseServicePage):
     template = "patterns/pages/service/service.html"
 
-    parent_page_types = ["ServicePage"]
+    parent_page_types = ["ServicePage", "propositions.PropositionPage"]
 
     @property
     def service(self):
+        proposition_page = PropositionPage.objects.ancestor_of(self).live().last()
         service_page = ServicePage.objects.ancestor_of(self).live().last()
```
ababic, line 329: We can add select_related("service") to both of these lines for a quick performance win

ababic (approved):

tbx/core/blocks.py
```diff
@@ -18,6 +22,61 @@
...
+    link_url = URLBlock(label="URL")
+    link_text = CharBlock()
+
+    class Meta:
+        label = "External link"
+        icon = "link"
+        value_class = LinkStructValue
+
+
+class LinkBlock(StreamBlock):
```
William-Blackie, line 61: Issue: Defining this as a streamblock and relying max_num to prevent more than one and then on indexing the first value feels a work around that does not need to exist.

Something like this removes this issue:

```
class LinkStructValue(StructValue):
    def url(self):
        external_url = self.get('external_link')
        internal_link = self.get('internal_link')
        return external_url or internal_link.url

class LinkBlock(blocks.StructBlock):
    internal_link = blocks.PageChooserBlock(required=False)
    external_link = blocks.URLBlock(required=False)

    class Meta:
        icon = 'link'
        template = 'utils/blocks/internal_external_link.html'

    def clean(self, value):
        errors = {}
        if value.get('internal_link') and value.get('external_link'):
            msg =  'Only one of these fields should be filled out.'
            errors['internal_link'] = ErrorList([
                msg
            ])
            errors['external_link'] = ErrorList([
                msg
            ])

        if not value.get('internal_link') and not value.get('external_link'):
            msg = 'You must fill out one of these fields.'
            errors['internal_link'] = ErrorList([
                msg
            ])
            errors['external_link'] = ErrorList([
                msg
            ])

        if errors:
            raise ValidationError('Validation error in InternalExternalLinkBlock', params=errors)

        return super(InternalExternalLinkBlock, self).clean(value)

    class Meta:
        value_class = LinkStructValue
```

William-Blackie: Please note that the FE templates using `link.value.o` needs changing to `link.url`

tbx/project_styleguide/templates/patterns/pages/careers/careers_page.html
```diff
@@ -9,7 +9,7 @@
 <div class="careers">
     {# Hero #}
     <div class="careers__hero">
-        {% include "patterns/molecules/hero/full_page_hero.html" with background_url=hero_image.url classes="hero__full-screen--short hero__full-screen--image-overlay" %}
+        {% include "patterns/molecules/hero/hero--full-screen.html" with background_url=hero_image.url classes="hero__full-screen--short hero__full-screen--image-overlay" %}
```
William-Blackie, line 12: Question: Why are we using BEM notation for filenames?

tbx/propositions/models.py
```diff
@@ -0,0 +1,128 @@
+from django.db import models
+
+from tbx.core.blocks import PageSectionStoryBlock
+from wagtail.admin.panels import FieldPanel, MultiFieldPanel
+from wagtail.fields import RichTextField, StreamField
+from wagtail.models import Page
+from wagtail.search import index
+
+INTRO_RICHTEXT_FEATURES = ["bold", "italic", "link", "document-link", "strikethrough"]
```
William-Blackie, line 9: Suggestion: Move this to it's own constant.py file, we don't have to but seperation is useful.

tbx/services/models.py
```diff
@@ -320,14 +322,28 @@ class ServicePageProcess(Orderable, BaseServicePageProcess):
...
+    parent_page_types = ["ServicePage", "propositions.PropositionPage"]
 
-    @property
+    @cached_property
     def service(self):
-        service_page = ServicePage.objects.ancestor_of(self).live().last()
-
-        if service_page:
-            return service_page.service
+        try:
```
William-Blackie, line 329: Issue: Why is this code repeated?

tbx/static_src/sass/components/_hero.scss
```diff
@@ -42,8 +42,108 @@
...
+            min-height: 700px;
+        }
+    }
+
+    &--image-overlay {
+        &::after {
+            content: '';
+            background-color: var(--color--black);
+            position: absolute;
+            top: 0;
```
William-Blackie, line 81: Suggestion: Use `inset: 0;` here instead of using four properties to set the position

William-Blackie (changes requested):
Hey all, there's a couple of things we should take care of before we can ship it. It should be pretty straight forward though!

tbx/services/models.py
William-Blackie: Weird, coming back to fix this, the diff has changed. Nothing to change.

tbx/core/blocks.py
William-Blackie: So, having picked this up to implement it, I know understand more about why these blocks were approached this way. 

It's not ideal and clearly is a work around due to how values are passed between blocks / child blocks (https://github.com/wagtail/wagtail/issues/5300#issuecomment-555708044). However, there's no way to keep the individual external and internal links in seperate classes without either combining them and writting custom validation in that inherited class or without writting a custom component to do so. 

Hopefully this can be fixed in a later iteration of Wagtail with the Streamfield API as is discussed on that ticket.

William-Blackie (approved):

PR #411, opened 2022-09-06
New streamfield template for client logos

helenb:
- Uses existing styles from the `organisms/client-block.html` template
- Adds links plus a hover effect
- Adds a title field for the link
- Adds comment that `organisms/client-block.html` could be removed when the service page is no longer needed.


tbx/project_styleguide/templates/patterns/molecules/streamfield/blocks/client-logo-block.html
```diff
@@ -0,0 +1,20 @@
+{% load wagtailimages_tags wagtailcore_tags %}
+<div class="client-block">
+    <div class="client-block__container">
+        {% if value.clients %}
```
chris-lawton, line 4: Suggestion: move the conditional to wrap the entire component so it won't output any empty divs if `value.clients` is false

PR #412, opened 2022-09-07
Add config for security headers

SharmaineLim:
https://projects.torchbox.com/projects/support-team/tickets/561

### When applied, this MR will...

set up the `Strict-Transport-Security` (HSTS), `Content-Security-Policy` (CSP), and `Permissions-Policy` headers.

To do:

- Permissions Policy headers

### Please provide detail on the technical changes, if relevant

- tweak some existing config:
    - Force `SECURE_SSL_REDIRECT` as true always
    - Set up a default value of 30 days for `SECURE_HSTS_SECONDS`
- use `django-csp` and environment variables for CSP headers, and
- use `django-permissions-policy` for Permissions Policy headers.
    - I've set most of the features as disabled because it doesn't look like we need them for wagtail.org (except `fullscreen` which I left out so that we can still use full screen with youtube videos)

### DIFF
--- a/tbx/settings/base.py
+++ b/tbx/settings/base.py
@@ -90,12 +90,15 @@ INSTALLED_APPS = [
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
+    "django_permissions_policy.PermissionsPolicyMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
+    # Clickjacking prevention. Default: X_FRAME_OPTIONS = 'DENY'
+    # See https://docs.djangoproject.com/en/dev/ref/clickjacking/#preventing-clickjacking
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "wagtail.contrib.redirects.middleware.RedirectMiddleware",
 ]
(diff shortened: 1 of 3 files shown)


b-ggs (approved):
LGTM!

tbx/settings/base.py
```diff
@@ -326,22 +329,91 @@
...
-    SECURE_BROWSER_XSS_FILTER = True
-
-if env.get("SECURE_CONTENT_TYPE_NOSNIFF", "true").lower().strip() == "true":
-    SECURE_CONTENT_TYPE_NOSNIFF = True
+# Do not use the `includeSubDomains` directive for HSTS. This needs to be prevented
+# because the apps are running on client domains (or our own for staging), that are
+# being used for other applications as well. We should therefore not impose any
+# restrictions on these unrelated applications.
+# https://docs.djangoproject.com/en/3.2/ref/settings/#secure-hsts-include-subdomains
+SECURE_HSTS_INCLUDE_SUBDOMAINS = False
```
William-Blackie, line 361: I feel like this might be a slight misunderstanding, as we set the sub-domains above this level (i.e CF) these won't filter down except for ones serve via this Wagtail instance. So the staging instances you've referenced above won't be affected.

tbx/settings/base.py
```diff
@@ -326,22 +329,91 @@
...
+
+# Permissions policy settings
+# Uses django-permissions-policy to return the header.
+# https://github.com/adamchainz/django-permissions-policy
+# The list of Chrome-supported features are in:
+# https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md
+PERMISSIONS_POLICY = {
+    "accelerometer": [],
+    "ambient-light-sensor": [],
+    "autoplay": [],
```
William-Blackie, line 402: I think we'd want to include `'self'` in autoplay due to the homepage including a `<video>` element with autoplay. Link to file [here](https://github.com/torchbox/wagtail-torchbox/blob/24eb9f5868523a23c7f982bc201887f15ff72590/tbx/project_styleguide/templates/patterns/pages/home/home_page.html) and MDN ref [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy/autoplay#:~:text=The%20HTTP%20Permissions%2DPolicy%20header,the%20Promise%20returned%20by%20HTMLMediaElement.).

tbx/settings/base.py
```diff
@@ -326,22 +329,91 @@
...
+    "camera": [],
+    "display-capture": [],
+    "document-domain": [],
+    "encrypted-media": [],
+    "geolocation": [],
+    "gyroscope": [],
+    "interest-cohort": [],
```
William-Blackie, line 409: I'm not sure if we'd want to keep this: https://web.dev/floc/

katdom13: Removed  "interest-cohort"

tbx/settings/base.py
katdom13: Added "self" to autoplay.

tbx/settings/base.py
katdom13: Thank you for this. I left that setting, but updated the comment to be like this one:
https://git.torchbox.com/internal/wagtail-kit/-/merge_requests/792/diffs#cfda3486b76cce823d1769ea08513c0e89f2ad9b_588_584

William-Blackie (approved):
Looks good, nice work @katdom13, feel free to merge to staging after rebasing / fixing the poetry conflicts.

PR #417, opened 2022-09-30
398 BE proposition logo block

victoriachan:
Ticket: https://projects.torchbox.com/projects/tbxcom/tickets/398

**IMPORTANT:** Before merging to `feature/389-BE-proposition-page` or the integration branch, delete `tbx/propositions/migrations/0002_clients_logo_block.py` and regenerate migration to avoid conflicts with other blocks that are developed concurrently.


tbx/project_styleguide/templates/patterns/molecules/streamfield/blocks/pullquote_block.html
```diff
@@ -18,7 +18,9 @@
         {% endif %}
     </cite>
 
-    {% if value.link %}
-        <a href="{% pageurl value.link %}" class="pullquote__link">{{ value.link_title|default:value.link.title}}</a>
+    {% with link=value.link.0.value %}
```
helenb, line 21: Query: I think this change and the one to pullquote_block.yaml might be a change from another ticket?

victoriachan:
@helenb  Ah sorry I got confused switching between the branches. I have removed the stray code from '396 BE Testimonial / quote block for proposition page' now. 

I've had to undo the previous commit and force push so that the removal will not then remove those files when these are all merged. So you may have problems pulling this branch again, if you ever need to. Best solution is to delete this branch from your local repo and pull again.

tbx/project_styleguide/templates/patterns/molecules/streamfield/blocks/pullquote_block.html
victoriachan: Removed now. Well spotted!

PR #423, opened 2022-10-24
Feature/layout fix

nicklee:
Updating proposition template css so it flows correctly

Screenshot 2022-10-24 at 14 41 21
Screenshot 2022-10-24 at 14 41 07

### DIFF
--- a/tbx/static_src/sass/components/_embed-cta.scss
+++ b/tbx/static_src/sass/components/_embed-cta.scss
@@ -23,6 +23,7 @@
     }
 
     &__button {
+        /* stylelint-disable-next-line declaration-no-important */
         color: var(--color--white) !important; // ovveride streamfield value
     }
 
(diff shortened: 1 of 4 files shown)


tbx/static_src/sass/components/_embed-cta.scss
```diff
@@ -22,6 +22,11 @@
         color: var(--color--grey-dark);
     }
 
+    &__button {
+        /* stylelint-disable-next-line declaration-no-important */
+        color: var(--color--white) !important; // ovveride streamfield value
```
siimonevans, line 27: Assuming this is the only way?

siimonevans (approved):

tbx/static_src/sass/components/_embed-cta.scss
nicklee: yea, without adding a stack of parent classes to increase specificity, which would also need a disable next line, this felt cleaner

PR #428, opened 2022-12-07
Add flight path

victoriachan:
Add flightpath as github actions: https://projects.torchbox.com/projects/support-team/tickets/641 

This adds github action that can be triggered by a button click (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). 

The button will only appear after this code is merged to the 'default' git branch (i.e. `master`). Once that is done, it can be triggered from the 'Actions' page on Github. Read more about [Manually running a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow?tool=webui).


b-ggs (approved):
LGTM!

.github/workflows/flightpath.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  run_flightpath:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install curl, jq
+        run: apt-get install -y curl jq
+      - name: Send POST request to FlightPath
+        run: |
+          export JOB_STATUS=$(curl -f -X POST -m 900
+          -H "Authorization: Token ${{ secrets.FLIGHTPATH_AUTH_KEY }}" -d "source_key=${{ secrets.DEPLOYMENT_KEY }}&destination_key=${{ secrets.DEPLOYMENT_KEY }}&copy_media=1&from_backup=1"
```
zerolab, line 20: I am still not OK with this as it these will appear in the action logs.

As mentioned in Slack, I'd very much prefer to move it to a python script

zerolab (changes requested):

.github/workflows/flightpath.yml
victoriachan: I've now moved it to python script.

.github/workflows/flightpath.yml
```diff
@@ -0,0 +1,31 @@
...
+#   - FLIGHTPATH_AUTH_KEY (from flightpath)
+#   - DEPLOYMENT_KEY (as set on corresponding heroku app env var)
+run-name: Copy ${{ secrets.HEROKU_APP_NAME_PRODUCTION }} to ${{ secrets.HEROKU_APP_NAME_STAGING }} by @${{ github.actor }}
+on: [workflow_dispatch]
+concurrency: staging_environment
+jobs:
+  run_flightpath:
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.9
```
zerolab, line 15: question: any reason not to use 3.11 and benefit from its performance improvements?

.github/workflows/flightpath.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests
+          pip install argparse
```
zerolab, line 23:
```suggestion
          python -m pip install requests
```

zerolab: argparse is part of the Python standard library as of Python 3.2

https://docs.python.org/3/library/argparse.html

tbx/core/utils/run_flightpath.py
```diff
@@ -0,0 +1,84 @@
...
+        # Don't accidentally run on prod!
+        raise ValueError(
+            "Destination cannot be the same as the source. Please specify a valid value for '--destination'."
+        )
+
+    # The following github vars may not exist if this is
+    # not called from github action, e.g. from command or shell
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+    repository = (os.environ.get("GITHUB_REPOSITORY"),)
+    run_id = (os.environ.get("GITHUB_RUN_ID"),)
```
zerolab, line 78: question: why are these defined as tuples?

tbx/core/utils/run_flightpath.py
```diff
@@ -0,0 +1,84 @@
...
+        args = {
+            "source": os.environ["HEROKU_APP_NAME_PRODUCTION"],
+            "flightpath_auth_key": os.environ["FLIGHTPATH_AUTH_KEY"],
+            "deployment_key": os.environ["DEPLOYMENT_KEY"],
+            "flightpath_url": os.environ["FLIGHTPATH_URL"],
+        }
+
+    except KeyError:
+        raise KeyError(
+            "You need the following environment variables to run flightpath: FLIGHTPATH_URL, HEROKU_APP_NAME_PRODUCTION, FLIGHTPATH_AUTH_KEY, DEPLOYMENT_KEY. This should be set on github secrets if running as github actions."
```
zerolab, line 24: nitpick: `github` -> `GitHub`, `github actions` -> `GitHub Actions` everywhere

tbx/core/utils/run_flightpath.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    # The following github vars may not exist if this is
+    # not called from github action, e.g. from command or shell
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+    repository = (os.environ.get("GITHUB_REPOSITORY"),)
+    run_id = (os.environ.get("GITHUB_RUN_ID"),)
+
+    print("Running flightpath...")
+    print(f"from Github repository: {repository}, run_id: {run_id}.")
```
zerolab, line 81: is the intetion for this only to run as part of GitHub Actions? if so, best to move it to `.github/`. Otherwise, perhaps tweak this note

zerolab (changes requested):

tbx/core/utils/run_flightpath.py
victoriachan: I was thinking of making the script reusable so it can be added to Kit. It can potentially be called from gitlab CI or  be used by fabric or django-admin commands.

.github/workflows/flightpath.yml
victoriachan: No reason not to! I've updated it now.

.github/workflows/flightpath.yml
```diff
@@ -0,0 +1,30 @@
...
+on: [workflow_dispatch]
+concurrency: staging_environment
+jobs:
+  run_flightpath:
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
```
zerolab, line 18: missed that earlier:

```suggestion
          python-version: '3.11'
```

tbx/core/utils/scripts/run_flightpath.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # To avoid accidentally running this script, a valid destination
+    # has to be manually specified rather than fetched from env var.
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--destination", help="Destination Heroku app name")
+    args = parser.parse_args()
+    if not args.destination:
+        raise ValueError("Please specify a destination with '--destination'.")
+    elif args.destination == flightpath_args["source"]:
+        # Don't accidentally run on prod!
```
zerolab, line 69: just nitpicking here: if one sets the wrong apps in the env vars it will happen anyway.

Been thinking.. since we're setting everything else as an env var, do we need to pass destination as an argument? why not read it from the env like the source/keys?

victoriachan: I wanted to make it harder to accidentally run it. Specifiying the destination makes it more deliberate. Sort of like Heroku making you retype the app now before it lets you delete it. Though it still won't be much help if someone just uses command line history to run it. What do you think? It is rather awkward.

.github/workflows/flightpath.yml
```diff
@@ -0,0 +1,30 @@
+name: copy_prod_to_staging
+# The following secrets are required on GitHub:
+#   - FLIGHTPATH_URL (URL for the flightpath call excluding trailing slash)
+#   - HEROKU_APP_NAME_PRODUCTION (source heroku app name)
+#   - HEROKU_APP_NAME_STAGING (destination heroku app name)
+#   - FLIGHTPATH_AUTH_KEY (from flightpath)
+#   - DEPLOYMENT_KEY (as set on corresponding heroku app env var)
+run-name: Copy ${{ secrets.HEROKU_APP_NAME_PRODUCTION }} to ${{ secrets.HEROKU_APP_NAME_STAGING }} by @${{ github.actor }}
+on: [workflow_dispatch]
```
nickmoreton, line 9: **Question**
I wonder if this could be adjusted so the manual command can be run from a staging pipeline, like we do in gitlab?

Then this could be tested by only pushing to staging.

victoriachan: I don't think so.

https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow  

“To run a workflow manually, the workflow must be configured to run on the workflow_dispatch event. To trigger the workflow_dispatch event, your workflow must be in the default branch.”

nickmoreton: Sorry I posted this on slack :) 

So is this where you mentioned you could changed the branch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch (scroll down a little to the Run workflow button)

And a little lower down it mentions limiting the branches: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches

I haven't explored this much so I may be missing something.

nickmoreton (approved):
The code looks to me to be ready to run the command and looks good. If it can only be run from the master branch that seems OK to me.

I'd need to put together a quick test repo to try out getting the button available on other branches because I don't completely understand the docs. 🙁

.github/workflows/flightpath.yml
victoriachan: This is what I found when testing on my fork: Once this is on master, admins will then see the 'copy prod to staging' option in the 'Actions' tab on Github website. When clicking on this, you can then choose to run the workflow using the script from a selected branch. Here you could choose staging if you like. But the option only appears if some form on this action is on the master/default branch first.

alxbridge:
Hi @victoriachan - you've addressed all the review feedback left by others, and this looks fine to me too.

If you wanted to test this before it's merged, I like the approach [suggested here](https://stackoverflow.com/a/65389878): creating a dummy workflow with the same name in `master`, causing the workflow to be registered in Github and allowing you to select your branch's version to run instead.

katdom13:
Hi @victoriachan ,
This also looks good to me. Thank you!

tbx/core/utils/scripts/run_flightpath.py
victoriachan: I have changed this now.

PR #440, opened 2023-01-17
Fix logic for conditional revision model

zerolab:
tests fail with 

```
raise FieldError("Cannot resolve keyword '%s' into field. "
django.core.exceptions.FieldError: Cannot resolve keyword 'page' into field. Choices are: approved_go_live_at, base_content_type, base_content_type_id, content, content_type, content_type_id, created_at, created_comments, id, object_id, object_str, submitted_for_moderation, task_states, user, user_id
```

on `PageRevision.objects.filter(page__content_type__in=...)`

Ref: https://github.com/torchbox/wagtail-torchbox/actions/runs/3938884880/jobs/6738095040#step:13:56

### DIFF
--- a/tbx/core/utils/migrations.py
+++ b/tbx/core/utils/migrations.py
@@ -23,19 +23,23 @@ def for_each_page_revision(*model_names):
         def wrapper(apps, schema_editor):
             ContentType = apps.get_model("contenttypes.ContentType")
 
-            try:
-                PageRevision = apps.get_model("wagtailcore.Revision")
-            except LookupError:
-                # In previous versions of Wagtail, this model was `PageRevision`
-                PageRevision = apps.get_model("wagtailcore.PageRevision")
-
             content_types = [
                 ContentType.objects.get_for_model(apps.get_model(model_name))
                 for model_name in model_names
             ]
-            revisions = PageRevision.objects.filter(
-                page__content_type__in=content_types
-            )
+
+            try:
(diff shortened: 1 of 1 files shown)


tbx/core/utils/migrations.py
```diff
@@ -23,21 +23,23 @@ def wrapper(fn):
...
             ContentType = apps.get_model("contenttypes.ContentType")
 
+            content_types = [
+                ContentType.objects.get_for_model(apps.get_model(model_name))
+                for model_name in model_names
+            ]
+
             try:
                 PageRevision = apps.get_model("wagtailcore.Revision")
+                revisions = PageRevision.objects.filter(content_type__in=content_types)
```
RealOrangeOne, line 33: Question: Is there a reason the `select_related` was removed here?

zerolab: yes. `Revision` is now generic. and `page` is a property - https://github.com/wagtail/wagtail/blob/main/wagtail/models/__init__.py#L2723

I suppose this could do with `select_related('content_object')`. Will add

RealOrangeOne: You probably need `.prefetch_related` for that, but I agree it's probably still useful.

zerolab: updated

zerolab: d'oh! it is `GenericForeignKey`, not `ForeignKey` 🙈

PR #441, opened 2023-01-17
Fix logic for conditional revision model (#440)

zerolab:
Follow up to #440, but for `main`


engineervix:
Hi @zerolab, is this still required on the master `branch`?

zerolab:
It is probably no longer needed, seeing how all tests are green

PR #442, opened 2023-01-26
Re-encode homepage showreel

thibaudcolas:
Re-encoded from the source `Showreel V5b.mp4` (1920x1080, 7.8MB).

This new version is slightly lower quality but not in a noticeable way, and 2x smaller. Encoding parameters for reference:

- H.264 via ffmpeg via Handbrake
- Video resized to 976x548 to match the display dimensions
- No audio channel (there was an empty one in the source)
- Constant quality with RF 27
- veryslow preset, tune none, profile main

I tried higher RF values but the quality wasn’t satisfactory, and a few different tunes but the difference wasn’t noticeable.


helenb:
Hey @thibaudcolas I don't have a local build set up to test this right now. The code changes look fine - just one query: does the smaller width and height of the video affect the page layout at all?

thibaudcolas:
@helenb no, the size I’ve re-encoded the video to is the maximum size it’s currently displayed at across the various viewports. The layout is still done with CSS anyway, so it’ll be the same before/after.

zerolab (approved):

PR #443, opened 2023-02-03
Fix client logo styling (#415)

jhancock532:
[Link to ticket on Codebase](https://projects.torchbox.com/projects/tbxcom/tickets/415)

The issue was caused due to an update to the client logo container element, which had been converted into a CSS grid system recently. The `client-item` children within the grid no longer need their width specified as a percentage, as the grid container will handle their width for them.

There was an issue with the streamfield version of the client logo block not resizing - I've addressed this by removing the fixed width of these logo elements.

I've updated the pattern library so you can preview the client logo block as the child of a streamfield within the propositions page. This is helpful as generic styles get applied to all the child elements of a streamfield container - if we're not careful, we'll miss the necessary resetting / overriding of these default streamfield styles while developing.

I've tweaked the padding on desktop slightly for a better final size of logos, see screenshots.

### Screenshots / Gifs

<details><summary>Screenshots</summary>
Screenshot 2023-02-03 at 11 17 05

Screenshot 2023-02-03 at 11 18 14

Screenshot 2023-02-03 at 11 18 53
</details>

[Gif of the client logo block in the pattern library](https://github.com/torchbox/wagtail-torchbox/files/10578872/resizing-pattern-library-template.gif.zip).

### DIFF
--- a/tbx/project_styleguide/templates/patterns/molecules/streamfield/streamfield-example-proposition-page/clients-section-body.html
+++ b/tbx/project_styleguide/templates/patterns/molecules/streamfield/streamfield-example-proposition-page/clients-section-body.html
@@ -1,0 +1,16 @@
+<div class="streamfield">
+
+    {% include "patterns/molecules/streamfield/blocks/client-logo-block.html"  %}
+
+    <div class="streamfield__quote quote-slider">
+        <svg class="quote-slider__icon" width="120" height="90" aria-hidden="true">
+            <use xlink:href="#quote-icon" />
+        </svg>
+        <div class="quote-slider__slider">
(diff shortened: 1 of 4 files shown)


helenb:
Hey @jhancock532 the styling changes here look fine. I'm not so keen on the example pattern library template added under `molecules/streamfield` as this is a bit non-standard. Could we not simply add the client logos to the `story_container` template so that they appear on page templates? https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/project_styleguide/templates/patterns/_pattern_library_only/streamfield/story_container.html

jhancock532:
Discussed over slack, happy to merge as is

PR #446, opened 2023-03-06
Add Hotjar tracking in CMS admin with insert_global_admin_js

thibaudcolas:
This is a trial of using Hotjar to track user actions in the CMS. We’re interested in this to inform changes to navigation across the site’s pages – we’ll collect data for a few weeks, and if this proves promising, endeavour to do the same type of analysis on a larger site.

To test this,

- Without any special configuration, this should do nothing.
- With a `ADMIN_HOTJAR_SITE_ID` environment variable, this should load the Hotjar tracking script.

For example:

```bash
export ADMIN_HOTJAR_SITE_ID=3367653
```

Hotjar site IDs are set up for specific domain names so this might not do anything aside from loading the script – and that’s ok.

---

Once deployed in Heroku, we will use config vars to track with the correct Hotjar account, and to turn this off once it’s no longer needed.

### DIFF
--- a/tbx/core/wagtail_hooks.py
+++ b/tbx/core/wagtail_hooks.py
@@ -1,6 +1,8 @@
+from django.conf import settings
 from django.core.files.storage import get_storage_class
 from django.shortcuts import redirect
 from django.utils.cache import add_never_cache_headers
+from django.utils.safestring import mark_safe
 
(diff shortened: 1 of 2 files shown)


helenb:
LGTM thanks @thibaudcolas

helenb (approved):

PR #448, opened 2023-03-07
Optional media on EmbedPlusCTABlock

engineervix:
This PR modifies the **EmbedPlusCTABlock** by ...

- making the `embed` field optional
- adding a new optional `image` field

... thereby allowing content editors the option of specifying either an image or embed.


victoriachan (approved):
Looks good! Works well when I tested locally too.

PR #450, opened 2023-03-07
#416 Filter 'See more case studies' by proposition

estyxx:


victoriachan (approved):
Looks good and works well. Assuming that all the pages using this (Sevice pages) have slugs that match a Service taxonomy term.

PR #453, opened 2023-03-07
#223 Service Process add external link

estyxx:


tbx/services/models.py
```diff
@@ -252,18 +252,27 @@ class Meta:
...
+        FieldPanel("external_link"),
+        FieldPanel("link_label"),
     ]
 
+    @property
+    def link(self):
+        if self.external_link:
+            return self.external_link
+        if self.page_link:
+            return self.page_link.url
```
victoriachan, line 274: Suggestion: `get_url` (which is what .url is behind the scenes) accepts a `request` argument, which can massively help performance. It might be better to do this switching in the template, so `{% page_url %}` can be used. 

Though in this case there is only one link (i.e. not looping through a bunch of pages), and the way you've done it does look cleaner. So it's up to you to weight performance vs code clarity.

victoriachan (changes requested):
Some tiny tweaks required for the template conditionals and an optional suggestion about using request with get_url.

tbx/project_styleguide/templates/patterns/molecules/process-item/process-item.html
```diff
@@ -4,8 +4,8 @@ <h2 class="process__item-title">{{ item.title }}</h2>
     <div class="process__item-desc">{{ item.description | richtext }}</div>
     {% if item.page_link %}
```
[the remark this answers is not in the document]
estyxx: nice spot!

tbx/services/models.py
victoriachan: Sorry if I wasn't clear before. What I meant is that providing the optional `request` argument [improves performance](https://github.com/wagtail/wagtail/blob/main/wagtail/models/__init__.py#L2128-L2131). It would be great if you could use `get_url(request)` but unfortunately `request` isn't available here to use in custom properties.

The alternative is to move this entire conditional logic to the template, so instead of using your convenient custom `item.link`, you do this  e.g.

```
{% if item.external_link %}
      <a href="{{ item.external_link }}">
{% elif item.page_link %}
      <a href="{% pageurl item.page_link %}">
{% endif %}
```

It does seem more clunky having this logic on the template. But it improves performance.

victoriachan (approved):
Hi Ester, looks good. I have replied regarding pageurl. It is just a suggestion, so happy to approve this as it is.

PR #461, opened 2023-03-09
Support: Upgrade to Wagtail to 4.2

victoriachan:
Ticket: https://torchbox.monday.com/boards/1130687432/pulses/1159465540

Upgrade Wagtail to 4.2. For completeness, also update the other packages used.

Considerations:

- There is a new migration for custom image. ([see](https://torchbox.monday.com/docs/1160010222?blockId=e1ff7bd8-d3fd-42ab-93ad-d69612506e8c))
- Comments within InlinePanel will be removed. ([see](https://torchbox.monday.com/docs/1160010222?blockId=c027acef-ec23-4758-b127-2824bbfc3758))


pyproject.toml
```diff
@@ -7,36 +7,36 @@ authors = ["Torchbox Ltd"]
...
 wagtail-accessibility = "^0.2.1"
 wagtailfontawesome = "^1.2.1"
-django-phonenumber-field = "^6.1.0"
+django-phonenumber-field = "^7.0.2"
 phonenumbers = "^8.12.48"
-wagtail-purge = {git = "https://github.com/nickmoreton/wagtail-purge", rev = "wagtail-4"}
-wagtail-webstories = {git = "https://github.com/nickmoreton/wagtail-webstories", rev = "allow-wagtail-4"}
+wagtail-purge = {git = "https://github.com/nickmoreton/wagtail-purge", rev = "support/wagtail42"}
+wagtail-webstories = {git = "https://github.com/nickmoreton/wagtail-webstories", rev = "support/wagtail42"}
 wagtail-markdown = "^0.10.0"
```
nickmoreton, line 30:
```suggestion
wagtail-markdown = {git = "https://github.com/nickmoreton/wagtail-markdown", rev = "support/wagtail42"}
```

I was seeing:
```
/venv/lib/python3.9/site-packages/wagtailmarkdown/blocks.py:4: RemovedInWagtail50Warning: Importing from wagtail.core.blocks is deprecated. Use wagtail.blocks instead. See https://docs.wagtail.org/en/stable/releases/3.0.html#changes-to-module-paths
  from wagtail.core.blocks import TextBlock
```
and there's a PR waiting for the maintainer to release

nickmoreton (commented):
This looks good except a couple of comments.

I was also seeing:
```
/venv/lib/python3.9/site-packages/wagtailfontawesome/wagtail_hooks.py:11: RemovedInWagtail50Warning: Importing from wagtail.core.hooks is deprecated. Use wagtail.hooks instead. See https://docs.wagtail.org/en/stable/releases/3.0.html#changes-to-module-paths
  from wagtail.core import hooks
```
so it looks like that package will need some attention. I have added to the packages board.

nickmoreton:
**Non blocking**: I also just noticed that: https://github.com/torchbox/wagtail-torchbox/blob/8e1433fe149b1fad7c941ce3f1d9d90577a66011/fabfile.py#LL437 the fab file run test django settings could be swapped to `"--settings=tbx.settings.test"`

pyproject.toml
victoriachan: Ah. Thanks for spotting this. 

Unfortunately when trying with this branch, the markdown block still doesn't seem to work properly. I am seeing the rich text buttons, which shouldn't be there, and the FE page doesn't render my markdown text correctly, my markdown text is just rendered as a single plain text line.

Wonder how much we use this and if we can just remove it. Will check.

victoriachan: I've updated the template now to use the markdown filter, as well as use your branch. It should now work.

victoriachan:
Hi @nickmoreton , Thanks for the review! I've fixed the problem with `wagtail-markdown`, and fabfile test settings, and have removed `wagtailfontawesome`.

nickmoreton:
This looks good to me.

I agree with removing the `wagtailfontawesome` as it's not serving much use.

PR #463, opened 2023-03-13
422 - FE styling issues | Proposition pages

cassidydvl:
[Ticket 422
](https://projects.torchbox.com/projects/tbxcom/tickets/422)

- fix header colors not reaching top of page
- quote spacing
- add streamfield section spacing
- theme--dark class overrides for proposition section (theme--dark seems to be used globally so was cautious about making changes to this class)

### DIFF
--- a/tbx/static_src/sass/components/_hero.scss
+++ b/tbx/static_src/sass/components/_hero.scss
@@ -118,9 +118,7 @@
             }
         }
         #{$root}__image-mask {
-            top: 280px;
-            // right: -400px;
-            // width: 75vw;
+            top: 140px;
         }
     }
 
--- a/tbx/static_src/sass/components/_key-points.scss
+++ b/tbx/static_src/sass/components/_key-points.scss
@@ -21,10 +21,12 @@
         }
     }
 
-    & &__link {
+    &__link {
         @include link-slide-right(4px, -5px);
-        &:hover {
-            border-bottom: 0;
+        .streamfield & {
+            &:hover {
+                border-bottom: 0;
+            }
         }
     }
 
--- a/tbx/static_src/sass/components/_proposition-section.scss
+++ b/tbx/static_src/sass/components/_proposition-section.scss
@@ -37,4 +37,15 @@
         color: var(--color--grey-dark);
         max-width: 760px;
     }
+
+    .streamfield {
+        margin: 0;
+        &:first-of-type {
+            margin-top: ($gutter * 3);
+
+            @include media-query(large) {
+                margin-top: ($gutter * 6);
+            }
+        }
+    }
 }
--- a/tbx/static_src/sass/components/_streamfield.scss
+++ b/tbx/static_src/sass/components/_streamfield.scss
@@ -212,15 +212,4 @@
             }
         }
     }
-
-    .proposition-section & {
-        margin: 0;
-        &:first-of-type{
-            margin-top: ($gutter * 3);
-
-            @include media-query(large) {
-                margin-top: ($gutter * 6);
-            }
-        }
-    }
 }


helenb:
Hi @cassidydvl I've changed this MR to point at `master` rather than `staging`. Once I've approved it, you can merge your branch into `staging` on your machine and push the `staging` branch to deploy it - leaving this MR in place for when it's been approved to go to production - does that make sense?

helenb (changes requested):
Hi @cassidydvl, some good fixes here, but I don't think overriding the theme css vars is quite working, so needs looking again. I think an approach of finding where the link text is white and overriding this for the proposition pages is the best one - I think this is usually determined by the `.streamfield` class. Alternatively we could see if there is a way to remove the `streamfield` class for the proposition pages.

tbx/static_src/sass/components/_theme.scss
```diff
@@ -68,11 +68,21 @@ Icon colour used for:
...
         --color--background: var(--color--dark-indigo);
-        --color--section-title: var(--color-coral);
+        --color--section-title: var(--color--coral);
         --color--header-icon-color: var(--color--lagoon);
         --color--get-in-touch-numbers: var(--color--dark-indigo);
         --color--header-links: var(--color--white);
         --color--icon: var(--color--lagoon);
+
+        & .proposition-section {
+            --color--link: var(--color--dark-indigo);
```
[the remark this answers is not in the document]
helenb: We discussed this in our call and Cassidy pointed out that the designs use the coral colour everywhere, so this does look correct. There's an issue with the hover state on the bottom borders not being quite right though - it might need to be a bit thicker.

Also, it might be best to override these css vars where we set the other styles for `.proposition-section` rather than nesting them here. Likewise the border hover style could be amended where the other `.streamfield` styles are defined.

Checking linting with `npm run lint`

tbx/static_src/sass/components/_theme.scss
```diff
@@ -68,11 +68,21 @@ Icon colour used for:
         --color--underline: var(--color--lagoon);
         --color--hover: var(--color--white);
         --color--background: var(--color--dark-indigo);
-        --color--section-title: var(--color-coral);
+        --color--section-title: var(--color--coral);
```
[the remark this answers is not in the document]
helenb: Design shows this as coral everywhere so we'll leave it and make sure Anna is aware.

tbx/static_src/sass/components/_hero.scss
```diff
@@ -118,6 +117,11 @@
                 padding-top: ($gutter * 3);
             }
         }
+        #{$root}__image-mask {
+            top: 280px;
```
cassidydvl, line 121: @helenb went this route just incase we want to bump up the image size like the design but some overflow issues were present with the image

tbx/static_src/sass/components/_key-points.scss
```diff
@@ -21,8 +21,11 @@
         }
     }
 
-    &__link {
+    & &__link {
```
cassidydvl, line 24: @helenb ran into some specificity issues, the .key-points parent class was being removed and streamfield styles were overriding without it, not sure this is the cleanest way to accomplish it

tbx/static_src/sass/components/_key-points.scss
```diff
@@ -21,8 +21,11 @@
...
         @include link-slide-right(4px, -5px);
+        &:hover {
+            border-bottom: 0;
+        }
```
cassidydvl, line 28: @helenb we could either remove the border or thicken it, i think the nice transition gets lost with the border

tbx/static_src/sass/components/_key-points.scss
helenb: Suggestion: change to

```
&__link {
        .streamfield & {
            @include link-slide-right(4px, -5px);
            &:hover {
                border-bottom: 0;
            }
        }
    }
```

That way it is clear what style you are overriding.

helenb (changes requested):
Thanks @cassidydvl - a couple more tweaks needed here. Could you also check the linting again - I think it is complaining about prettier this time. You can fix that by running `npm run format` in the terminal, but it would be great to get your editor set up to run prettier on save - Steve should be able to help you with that as I'm not around later.

tbx/static_src/sass/components/_hero.scss
```diff
@@ -118,6 +117,11 @@
...
+            // right: -400px;
```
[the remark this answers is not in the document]
cassidydvl: Apologies! I meant to have a chat about this and demonstrate. The commented code:
```
// right: -400px;
// width: 75vw;
```
is more consistent with the size of the image in the design but there is overlapping issues and overflow: visible set on [line 90](https://github.com/torchbox/wagtail-torchbox/pull/463/commits/a4d14d00c7a139d39d3a71b3465710637bc607ab#diff-46174d8126148a2383e0533f71f0795d90b7396391cae50630086f4952d5d0ceR90) so I'm assuming this was intentional? What do you think? Leave the size as is and position it to not overlap as suggested above?

tbx/static_src/sass/components/_key-points.scss
cassidydvl: Ah much better! Thank you!

PR #464, opened 2023-03-14
#416 Fix filter-by link in SubServicePage

estyxx:


tbx/services/models.py
```diff
@@ -347,6 +351,12 @@ class SubServicePage(BaseServicePage):
         FieldPanel("content", heading="Content"),
     ]
 
+    @property
+    def case_studies_filter_by(self):
+        # Filter with the parent's slug. Example:
+        # `/digital-products/product-strategy/` use filter by `digital-products`
+        return self.url.split("/")[1]
```
victoriachan, line 358: I find URL string splitting prone to errors, but you won't need this anymore if using `self.service.slug` (in `case_studies_filter_by` as defined in `BaseServicePage`). So this definition of `case_studies_filter_by` can be removed.

tbx/services/models.py
```diff
@@ -172,6 +172,10 @@ def get_featured_case_studies(self):
             if f.case_study
         ]
 
+    @property
+    def case_studies_filter_by(self):
+        return self.slug
+
```
victoriachan, line 178: Ah, I missed it in my earlier code review that instead of slug, you could use the `service` related field, which would give better results as it links directly to the service taxonomy and will match the filter on the WorkIndexPage. 

Also since the subservicepage has a `service` property to mirror the related field, you will only need to define this at `BaseServicePage`.

E.g.

```suggestion
    @property
    def case_studies_filter_by(self):
        if self.service:
            return self.service.slug
        
        # If no service defined, don't filter by anything
        return ""
```

victoriachan (changes requested):
Hi @estyxx, while reviewing this I realised it would be much better to use the `service` related field's slug instead of page slug. What do you think?

tbx/services/models.py
estyxx: Yes! I knew must be a better way! Thank you!

victoriachan (approved):

PR #465, opened 2023-03-14
Fix subnav z-index

estyxx:
The subnav menus was appearing under the hero image. 

I tried to put to `subnav` z-index the `nav` value (50) and `__image-mask` z-index `header` (30) but seems not 
working... 

Some page example:
I verified in other pages in this way the subnav is displayed correctly
CleanShot 2023-03-15 at 10 41 08

CleanShot 2023-03-15 at 10 42 23
CleanShot 2023-03-15 at 10 42 13

### DIFF
--- a/tbx/static_src/sass/components/_hero.scss
+++ b/tbx/static_src/sass/components/_hero.scss
@@ -126,7 +126,7 @@
     }
 
     &__image-mask {
-        @include z-index(header); // above in page nav to allow for overlap
+        @include z-index(zero);
         display: none;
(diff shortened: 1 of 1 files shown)


victoriachan (commented):
Hi @estyxx, your fix works for the issue, but I am concerned that according to the comment, @helenb had explicitly set this as 'above in page nav to allow for overlap'. Making it `zero` would have made it below in page nav (25). 

Have you checked with Helen if this would not break its relationship with the in page nav?

PR #469, opened 2023-03-16
Add filter_by= to blog links

estyxx:


victoriachan (changes requested):
Well done for refactoring the case studies filter for the Service pages to work with this! Now it got me to thinking we might have to refactor the way the filter is propagated to the `see-more.html` include. 

In your MR, the `related-content.html` include breaks in `ValuesPage` as it does not have `page.filter_by` defined.

We probably shouldn't have assumed page context variables in the includes. Instead, we can pass it as include variables. 

We can also make `filter_by` an optional variable in the see-more.html, related-content.html, and latest-post.html includes instead of passing it in with the link variable? 

e.g. in `see-more.html`
```
{% if filter_by %}
    <a class="see-more__link" tabindex="0" href="{{ link|add:'?filter_by='|add:filter_by }}">{{ text }}</a>
{% else %}
    <a class="see-more__link" tabindex="0" href="{{ link }}">{{ text }}</a>
{% endif %}
 ```

in intermediate includes, such as `latest-post.html` and `related-content.html`
```
{% include "patterns/atoms/see-more/see-more.html" with text=... link=... filter_by=filter_by %}
```

And then in the page template, e.g. `service.html`
```
{% if page.case_studies_section_title %}
    {% include "patterns/organisms/related-content/related-content.html" with id=... title=...  filter_by=page.filter_by %}
{% endif %}

{% if page.blogs_section_title %}
    {% include "patterns/organisms/latest-posts/latest-posts.html" with classes=... filter_by=page.filter_by %}
{% endif %}
```

Let me know if this is unclear!

File: tbx/project_styleguide/templates/patterns/organisms/latest-posts/latest-posts.html [hunk 1]
```diff
@@ -18,6 +18,6 @@ <h2 class="blog-listing__heading">What our team says</h2>
             {% endfor %}
         </div>
         {% pageurl blog_index_page as blog_index_url %}
-        {% include "patterns/atoms/see-more/see-more.html" with text="See more posts" link=blog_index_url %}
+        {% include "patterns/atoms/see-more/see-more.html" with text="See more posts"  link=work_index_page.url filter_by=filter_by %}
```
victoriachan, line 21: Typo here?
```suggestion
        {% include "patterns/atoms/see-more/see-more.html" with text="See more posts"  link=blog_index_url filter_by=filter_by %}
```

victoriachan (changes requested):
Thanks for refactoring. Sorry for all the extra work. We're nearly there! Just a typo to correct on latest-posts.html and also the include on the [Careers page  template](https://github.com/torchbox/wagtail-torchbox/blob/417-blog-link-add-filter-by/tbx/project_styleguide/templates/patterns/pages/careers/careers_page.html#L53) needed `filter_by=page.filter_by` added.

estyxx:
done! thanks!

PR #426, opened 2022-12-06
Add birdbath

victoriachan:
Add birdbath to the project as part of the progress towards implementing flightpath: https://projects.torchbox.com/projects/support-team/tickets/641

This should be quite straigtforward as the site doesn't have any non-torchbox users or particularly sensitive data.

PR #471, opened 2023-03-17
#426 proposition subpage

estyxx:
This PR 

- Add `external_link` to EmbedPlusCTABlock
- Don't show client-book's title if there aren't any client
- Add embed URL video to Processes panel in service page


tbx/project_styleguide/templates/patterns/organisms/process-block/process-block.html
```diff
@@ -9,6 +11,12 @@
             </div>
         {% endif %}
 
+        {% if processes_section_embed_url %}
+                <div class="process__embed">
+                    {% embed processes_section_embed_url %}
```
victoriachan, line 16: Question: Do we need Frontend dev to look at this to ensure the formatting, size, etc. works across different widths?

estyxx: Probably... actually the style .process__embed doesn't exist and I was about to copy past from `_embed-cta.scss`...

estyxx: yeah I just fixed a bit the styles, took from [there](https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/static_src/sass/components/_streamfield.scss#L30), it's nicer now but would be great to have a FE dev opinion :D

estyxx: CleanShot 2023-03-20 at 12 50 27

victoriachan: Yes, you might want to ask a FE dev. Unfortunately I can't help with this.

PR #478, opened 2023-04-17
Update Page models to include Wagtail Page's content_panels

victoriachan:
Ticket: https://projects.torchbox.com/projects/tbxcom/tickets/424

Many of the page models which are inheriting from wagtail's Page models are declaring their own `content_panels` and manually including the field panel for the pages' `title` field instead of using Wagtail core's Page content_panels. 

This change prepends the customn `content_panels` with Page.content_panels and also remove the custom title field panel so we can use Wagtail's better default title field.


estyxx (approved):
LGTM!

PR #484, opened 2023-05-11
Add a Call To Action streamfield field to almost every page

victoriachan:
Ticket: https://torchbox.monday.com/boards/1192293412/pulses/1192308132

Lisa wants to be able to add a 'call to action' content to every page that doesn't already allow that. Ideally she wants to be able to place it anywhere on the page. But that will require converting the existing fields to streamfield blocks. We have decided to do that as a separate ticket, as it would involve a lot of change and possibly data migrations.


estyxx (approved):
LGTM

PR #485, opened 2023-06-08
Fix Poetry installer in docker and also CI.

victoriachan:
At first Docker was complaining about mismatched installer SHA when building. And then the CI was failing due to Gr1n/setup-poetry@v7 being 404. So I have upgraded it to v8 (see https://github.com/Gr1N/setup-poetry).

I have also changed Poetry to be installed using Pip, like we do these days.

NOTE: Sorry, also found I was previously using this similar branch name (this branch `fix/update-poetry-sha` VS the old branch `fix/update_poetry_sha`) to update the SHA and then pushed it to staging and I have somehow lost track of it. I have no recollection of it at all! Treat this as a new branch and ignore that old one.


nickmoreton (approved):
Looks good to me.

**nitpik** Unless you want to update the poetry version to 1.5.1

nickmoreton (approved):
Looks good to me.

**nitpik** Unless you want to update the poetry version to 1.5.1

victoriachan:
Thanks, @nickmoreton I've updated this to Poetry 1.5.1 now. Will proceed to put it on staging.

PR #487, opened 2023-06-13
Fix proposition header image

albinazs:
https://torchbox.monday.com/boards/1192293412/views/4019870/pulses/1193034463

### **Changes included in this MR:**

- ensures the proposition header image does not expand out of its container
- adds a new breakpoint to make sure the header image looks good on wider screens (QHD +)

Note: I was thinking to add an additional breakpoint for screens with QHD < resolution < 4K and set it to 30vh, but that might be an overkill. Please let me know what you think.

### **To test**

On a local build visit the proposition page and see how the image behaves at various resolutions

### **Screenshots**

1920*1200

Screen Shot 2023-06-13 at 16 03 18 PM

2560*1440 (QHD)

Screen Shot 2023-06-13 at 16 04 18 PM

3840*1260 (4K)

Screen Shot 2023-06-13 at 16 04 46 PM

5120*2160 (5K)

Screen Shot 2023-06-13 at 16 05 04 PM

### DIFF
--- a/tbx/static_src/sass/abstracts/_variables.scss
+++ b/tbx/static_src/sass/abstracts/_variables.scss
@@ -96,7 +96,8 @@ $breakpoints: (
     'large' '(min-width: 1023px)',
     'menu-break' '(min-width: 1255px)',
     'menu-break-larger' '(min-width: 1400px)',
-    'x-large' '(min-width: 1919px)'
+    'x-large' '(min-width: 1919px)',
+    'xx-large' '(min-width: 2559px)'
 );
 
 // --------------------------------- Grid Dimensions --------------------------------------
--- a/tbx/static_src/sass/abstracts/_variables.scss
+++ b/tbx/static_src/sass/abstracts/_variables.scss
@@ -96,8 +96,7 @@ $breakpoints: (
     'large' '(min-width: 1023px)',
     'menu-break' '(min-width: 1255px)',
     'menu-break-larger' '(min-width: 1400px)',
-    'x-large' '(min-width: 1919px)',
-    'xx-large' '(min-width: 2559px)'
+    'x-large' '(min-width: 1919px)'
 );
 
 // --------------------------------- Grid Dimensions --------------------------------------
--- a/tbx/static_src/sass/components/_hero.scss
+++ b/tbx/static_src/sass/components/_hero.scss
@@ -117,17 +117,13 @@
                 padding-top: ($gutter * 3);
             }
         }
-        #{$root}__image-mask {
-            top: 140px;
-        }
     }
 
+    // The following code for image & image-mask is only used on proposition pages currently
     &__image {
         display: block;
         margin-top: -25px;
-        max-width: 100%;
         max-height: 50vh;
-        object-fit: cover;
 
         @include media-query(xx-large) {
             max-height: 20vh;
@@ -138,7 +134,7 @@
         @include z-index(zero);
         display: none;
         position: absolute;
-        top: -60px;
+        top: 140px;
         right: -50px;
         width: 45vw;
         height: auto;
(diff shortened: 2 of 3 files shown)


tbx/static_src/sass/components/_hero.scss
```diff
@@ -125,7 +125,13 @@
     &__image {
         display: block;
         margin-top: -25px;
-        width: 100%;
+        max-width: 100%;
```
helenb, line 128: Hi @albinazs I was going to query this and point out it will affect images on other hero elements, not just the proposition page - but having checked, the `hero__image` and `hero__image-mask` classes are only used in the proposition hero. This is a bit confusing, as the `hero__image-mask` class currently has a base style which is overridden for the `--proposition` variant styles.
Checking the `hero` class more widely:
- The careers page uses a full-screen hero which is a background image, so works differently.
- The service page and values page include the hero but don't have an image.
- The sign up form uses a `hero` class but not `hero__image`. 

So this all confirms that we don't need an override inside the `--proposition` variant for either `hero__image` or `hero__image-mask` - the style variation for the `image-mask` that is inside the `--proposition` variant could be pulled out.

helenb: You could maybe add a comment to the effect that these are only used on proposition pages currently

tbx/static_src/sass/abstracts/_variables.scss
```diff
@@ -95,7 +95,8 @@ $breakpoints: (
     'medium' '(min-width: 599px)',
     'large' '(min-width: 1023px)',
     'menu-break' '(min-width: 1255px)',
-    'x-large' '(min-width: 1919px)'
+    'x-large' '(min-width: 1919px)',
+    'xx-large' '(min-width: 2559px)'
```
helenb, line 99: I also added an extra breakpoint in my menu changes but it was a smaller one between `menu-break` and `x-large` so this is fine

helenb (commented):
HI @albinazs this looks good. I have suggested a small tweak to tidy up the  `hero_image-mask` styles. I wonder if we should also look at the resolution of the image used in the hero, whether it needs to be so large, and whether a fill might be more suitable than a `width` to closer match the size it will render

tbx/static_src/sass/abstracts/_variables.scss
helenb: > HI @albinazs this looks good. I have suggested a small tweak to tidy up the `hero_image-mask` styles. I wonder if we should also look at the resolution of the image used in the hero, whether it needs to be so large, and whether a fill might be more suitable than a `width` to closer match the size it will render

Oh and to answer your question in the description I do think a further break would be overkill.

helenb:
To answer your query in the description I think a further breakpoint would be overkill.

albinazs:
thank you @helenb!

I've made a cleanup inspired by your comments:
- pulled out redundant --proposition variant
- removed unnecessary declarations from my solution 

I've also tested using smaller images, they certainly shouldn't be as large as some of those on staging (especially after the changes we've made).

helenb:
Hi @albinazs this is looking good now. Regarding the images, I was talking about the resize of the image in the wagtail template: https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/project_styleguide/templates/patterns/pages/proposition/proposition.html#L6 - I wondered if this needed to be as wide and whether using a `fill` image resize rather than a `width` might be better - but I think that for wider screens it may need to be bigger the way it is coded right now - just seems a bit mad to have such a large image for quite a small space that it appears. I think let's get this merged now and then we could potentially look at this as part of another ticket.

helenb:
I've created a new ticket for the image size

helenb:
@albinazs Lisa and Anna felt the image had got too small at larger screen sizes, so this is an alternative fix. I've also moved my hero image rendition resize into this MR, and closed https://github.com/torchbox/wagtail-torchbox/pull/492

PR #488, opened 2023-06-14
Support external links in navigation

engineervix:
https://torchbox.monday.com/boards/1192293412/pulses/1206552545

This MR adds external link support for **primary navigation** and **footer links**

This is done by modifying `tbx.navigation.models.LinkBlock`.

In addition,

- the first `n` items in primary nav is bumped from 3 to 4
- [wagtail-factories](https://github.com/wagtail/wagtail-factories) is installed to help with tests
- the `LinkColumnWithHeader` block is removed as it is not used anywhere

<details>
  <summary>Screenshots</summary>

  Screenshot from 2023-06-14 09-45-22

  image

  image

  image

  image

</details>


tbx/project_styleguide/templates/patterns/molecules/navigation/footer-top-links.html
```diff
@@ -2,7 +2,7 @@
 <div class="footer__top">
```
helenb, line 2: I don't think this tag is used anywhere on the site. I wonder if we could remove the footer top links completely to tidy things up?

helenb: Looks like it was removed here: https://github.com/torchbox/wagtail-torchbox/commit/072b38fb58c4a3ad02ee5

engineervix: Good spot!

We should be able to do this by

- [ ] [removing `footertoplinks` inclusion tag](https://github.com/torchbox/wagtail-torchbox/blob/56bab81d886eb7262db5c306ed85b43d26537020/tbx/navigation/templatetags/navigation_tags.py#L62-L73) from `tbx/navigation/templatetags/navigation_tags.py`
- [ ] [removing the pattern library override](https://github.com/torchbox/wagtail-torchbox/blob/56bab81d886eb7262db5c306ed85b43d26537020/tbx/project_styleguide/templatetags/navigation_tags.py#L7) for the `footertoplinks` tag.
- [ ] deleting the `tbx/project_styleguide/templates/patterns/molecules/navigation/footer-top-links.html` template
- [ ] [removing `footer_top_links` from `NavigationSettings`](https://github.com/torchbox/wagtail-torchbox/blob/56bab81d886eb7262db5c306ed85b43d26537020/tbx/navigation/models.py#L62-L67), and the [corresponding `FieldPanel`](https://github.com/torchbox/wagtail-torchbox/blob/56bab81d886eb7262db5c306ed85b43d26537020/tbx/navigation/models.py#L79)
- [ ] generating migrations for `NavigationSettings`

Do you think we should do it here or as a separate PR?

helenb: I think it would make sense to do it here. That list looks good to me.

engineervix: I have removed footer top links via 5504b2ac

victoriachan (approved):
Note: I didn't review the .scss or .js changes as they are Frontend changes.

@engineervix I'm happy with your BE changes, and the approach you have taken. It has the least disturbance to the existing data, and does not need custom migration to move the navigation data to a new system.

You probably have considered this already, but just want to point out an alternative would be to replace the existing `LinkBlock` from a StructBlock to a StreamBlock which then nests a further external link and internal link blocks. (see https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/core/blocks.py#L64, could even reuse this) The benefit of this approach is that you don't have to do your custom clean validation. 

But I agree that your approach is a better as it does not change the existing data structure or FE code much, even if it requires custom validation. Well done for adding tests for those too!

PR #477, opened 2023-04-11
Ticket #429 - Proposition and sub-proposition pages | remove 'dark with transparent header theme'

cassidydvl:
[Ticket 429](https://projects.torchbox.com/projects/tbxcom/tickets/429)

- removes unused Dark with transparent header option from theme choices for proposition pages
- request was to remove from sub page types as well but they're not only sub-proposition pages ([see subpages](https://github.com/torchbox/wagtail-torchbox/blob/master/tbx/propositions/models.py#L15)) so the option has been left on those page types just in case the dark--transparent class is being used there

PR #490, opened 2023-06-15
Theme fixes for proposition pages (and elsewhere

helenb:
Theme fixes for proposition pages (and elsewhere), and remove the option to add the dark transparent theme for proposition pages

- Ticket: https://torchbox.monday.com/boards/1192293412/views/4019870/pulses/1193036756?term=image
- I have not followed a design for this

Overview of changes
- Removes the slightly crude overrides of theme colours that were previously done for all proposition pages, and instead adds some extra theme variables to handle the fact that the dark theme has links that sit both on the dark header area, and on the white body of the page.
- Removes the option to set a proposition page with the 'dark transparent' theme, as this has never worked.

The main aim of this ticket was to ensure that the 'dark' theme used the light blue colours for accents - see this preview of an unpublished service page on the live site to see how this previously looked for service pages: https://torchbox.com/admin/pages/1196/edit/preview/?mode=

In testing though I found a number of issues with different components in different themes which I have addressed as part of this ticket.

To test:
The pattern library page (http://localhost:8000/pattern-library/render-pattern/patterns/pages/proposition/proposition.html) uses the light theme, and does not show all the elements, so it is best to create a new wagtail proposition page (or edit an existing one), and test the following items:
- Links in the intro field (in the hero) for all themes (these have changed a bit in order to create a hover effect)
- In page nav links in the hero and sticky header in all themes
- Links in intro fields for 'our services', 'our clients' etc
- Key points headings in all themes
- CTA without an embed in all themes, with a span tag to change the colour
- Check the hover state on the CTA button doesn't add a pale blue bottom border
- CTA *with* an embed in all themes, with a span tag to change the colour
- Check the hover state on the CTA with embed button doesn't add a pale blue bottom border


helenb:
@engineervix would you mind reviewing my model change here? We should maybe create a data migration in case any proposition pages have the 'dark transparent' theme set, but I think it is unlikely any would as it doesn't work. If you think it needs a migration, it would be great if you could give me some advice on how to go about it.

engineervix:
> @engineervix would you mind reviewing my model change here? We should maybe create a data migration in case any proposition pages have the 'dark transparent' theme set, but I think it is unlikely any would as it doesn't work. If you think it needs a migration, it would be great if you could give me some advice on how to go about it.

Hi @helenb I've had a look at the model change and I think a data migration will be necessary, since existing Pages with the "dark--transparent" theme will not be automatically updated, and this may result in `ValueError`s.

Here's how I'd go about it:

1. Create a new data migration before removing the option.

  ```shell
  python manage.py makemigrations propositions --name data_migration_proposition_theme --empty 
  ```

  ```python
  from django.db import migrations
  
  
  def update_dark_transparent_theme(apps, schema_editor):
      Page = apps.get_model('propositions', 'PropositionPage')
      Page.objects.filter(theme='dark--transparent').update(theme='light')
  
  
  class Migration(migrations.Migration):
  
      dependencies = [
          ('propositions', '0009_auto_20230503_1527'),
      ]
  
      operations = [
          migrations.RunPython(update_dark_transparent_theme),
      ]
  ```

2. Run the migration to update the existing data.

  ```shell
  python manage.py migrate propositions
  ```

3. Remove the option from the model.
4. Generate a new migration after removing the option

helenb:
> > @engineervix would you mind reviewing my model change here? We should maybe create a data migration in case any proposition pages have the 'dark transparent' theme set, but I think it is unlikely any would as it doesn't work. If you think it needs a migration, it would be great if you could give me some advice on how to go about it.
> 
> > Note
> > Just realized that this seems to have already been done in #477

Ah I didn't see this comment in time and I've now pushed it to this update. I think we could actually close 477 now and just have the change here.

helenb:
@engineervix if #477 was already on staging will this not work when deployed as it assumes the dark theme is in place when running the data migration?

tbx/static_src/sass/components/_key-points.scss
```diff
@@ -28,6 +28,11 @@
                 border-bottom: 0;
             }
         }
+
+        &::before {
+            //overde underline colour in mixin
```
albinazs, line 33: override?

albinazs (approved):
@helenb, looks good to me!

tbx/static_src/sass/components/_key-points.scss
helenb: oops, will amend!

PR #492, opened 2023-06-20
Reduce the hero image size on proposition pages

helenb:
Ticket: https://torchbox.monday.com/boards/1192293412/views/4019870/pulses/1193034463

This builds on the work done in  https://github.com/torchbox/wagtail-torchbox/pull/487 so includes the code changes from there.

My only addition is to change the image tag to make the rendition used in the template smaller - it also now uses a crop to get an image that is the correct aspect ratio for the mask surrounding it (although with images where the focus is to the right, the focus may be hidden).

I haven't included a screenshot, as this should hopefully not make any visual changes.

### DIFF
--- a/tbx/project_styleguide/templates/patterns/molecules/hero/hero--proposition.html
+++ b/tbx/project_styleguide/templates/patterns/molecules/hero/hero--proposition.html
@@ -9,7 +9,7 @@
         </div>
         {% if image %}
             <div class="hero__image-mask">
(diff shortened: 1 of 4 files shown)


helenb:
I'm going to close this as the styling for https://github.com/torchbox/wagtail-torchbox/pull/487 has changed, and now incorporates a new header image size. This will overwrite what was on staging for this MR.

PR #494, opened 2023-06-21
Add a new streamfield block for the mailchimp embed

helenb:
@albinazs and @engineervix this creates the new block template for mailchimp. Albina, you can uses whatever sensible class you like for the template, and you don't need to share the `raw-html-block` styles if that isn't appropriate.


helenb:
NB the failed pipleine is from `npm run format` which is coming from Albina's code - her MR is only draft, I just based this one of it as I knew I wanted to merge to it.

engineervix (approved):
Thanks @helenb this looks good :+1:

PR #495, opened 2023-06-23
Data migration and update to service and sub-service page themes

helenb:
Ticket: https://torchbox.monday.com/boards/1192293412/views/4019870

This removes the 'dark transparent header' theme from service and sub-service pages. (Note that the ticket refers to sub-proposition pages, but there is no page with this type defined)

Updates any existing pages using the 'dark transparent header' theme to use the light theme, then removes that option for service and sub-service pages.

Before:
Screenshot 2023-06-23 at 10 45 52

After:
Screenshot 2023-06-23 at 10 46 01

Note that it was decided to apply this to service page types as well as sub-service page types after the first work had been done - one data migration would probably have been preferable, but I couldn't reverse the first migration that I ran.


engineervix (approved):
Thanks @helenb changes look good :+1:

engineervix:
I've added a commit which combines the data migrations into one module.

PR #496, opened 2023-06-26
Add wagtailmedia and custom video block

zerolab:
This PR:

- adds wagtailmedia, and a custom VideoBlock to StoryBlock. The block provides the option to:
   - autoplay which will autoplay, loop and mute the video
   - use the original width (up to the content width). Useful for when upscaling is not desirable
- adds some styling for the new vide block
- does some tidy up and bumps the isort version as it would fail to install locally for me.


tbx/static_src/sass/components/_streamfield.scss
```diff
@@ -53,6 +53,25 @@
...
+            width: 100%;
+            height: auto;
+        }
+
+        &--original video {
+            width: initial;
+            max-width: 100%;
+        }
+    }
+
```
zerolab, line 74: These are the styling changes for the new block. Decided it needs its own class rather than bundling it into the embed one above.

Notable here: by default the video element will be 100% of the content width. But for the original, it keeps its actual width (up to 100% of the content width) and is centered

tbx/core/blocks.py
```diff
@@ -211,6 +212,26 @@ class CTABlock(StructBlock):
     link = LinkBlock()
 
 
+class VideoBlock(StructBlock):
```
zerolab, line 215: the StreamField block definition.

engineervix (commented):
Thanks @zerolab, changes look good to me :+1: 

@helenb would you please take a look at the FE?

PR #502, opened 2023-06-29
Add CTA links to the Clients logos on hover & focus

albinazs:
https://torchbox.monday.com/boards/1192293412/pulses/1193604342

This PR adds CTA links on hovering Clients' logos:
- links are visible only on desktop screens (as there seems to be no use in hovering on tablets and mobiles)
- links texts are aria-hidden to avoid repeating it for screen reader users
- the grid is adjusted to accommodate the new content and prevent text-wrap
And this PR also updates the hover effect for the Footer logos to keep consistency.

A couple of related issues were noticed:
- Wharton interactive logo is noticeably poor quality
- logos on the Wagtail proposition page are all different sizes which makes the grid a bit less clean (compared to the Digital products proposition page)

### How to Test

On a local build, go to the Wagtail proposition page and test the following:
**Client logos section**
- how the grid adjusts on different screens
- how CTA link and after element look on different desktop screens (text should not wrap and after element should be positioned well)
- how CTA link and after element look on different color themes
- how smooth is scaling animation preventing the image from slightly shifting

**Footer logos section**
- how is the animation consistent with the Client's logo section

### Screenshots

<details>

image

image

</details>

### MR Checklist

### DIFF
--- a/tbx/static_src/sass/components/_client-block.scss
+++ b/tbx/static_src/sass/components/_client-block.scss
@@ -32,7 +32,7 @@
         }
 
         @include media-query(x-large) {
-            column-gap: 130px;
+            column-gap: 60px;
         }
 
         &--spaced {
--- a/tbx/static_src/sass/components/_client-block.scss
+++ b/tbx/static_src/sass/components/_client-block.scss
@@ -24,8 +24,11 @@
         }
 
         @include media-query(large) {
+            grid-template-columns: repeat(4, 1fr);
+        }
+
+        @include media-query(menu-break-larger) {
             grid-template-columns: repeat(6, 1fr);
-            column-gap: 20px;
         }
 
         @include media-query(x-large) {
--- a/tbx/static_src/sass/components/_client-item.scss
+++ b/tbx/static_src/sass/components/_client-item.scss
@@ -18,17 +18,25 @@
 
     &__image {
         width: 100%;
-        height: 100%;
-        transition: box-shadow $transition-quick;
+        align-self: center;
+        // for smooth scaling without shifting the image
+        transform: scale3d(1, 1, 1);
+        transition: transform $transition-quick;
 
         #{$root}__link:hover & {
-            box-shadow: 2px 2px 10px var(--color--grey);
+            transform: scale3d(1.05, 1.05, 1.05);
         }
     }
 
     &__link {
-        display: block;
+        height: 100%;
+        display: grid;
+        grid-template-rows: 1fr;
         border-bottom: 2px solid transparent; // prevent layout shift
+
+        @include media-query(large) {
+            grid-template-rows: 2fr 1fr;
+        }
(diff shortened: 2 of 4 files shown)


tbx/static_src/sass/components/_client-item.scss
```diff
@@ -39,20 +47,70 @@
...
 
             @include media-query(large) {
-                padding: 15px;
+                display: flex;
+            }
+
+            &::after {
+                content: '';
+                display: block;
+                position: absolute;
```
helenb, line 99: I would suggest changing this to `position: relative` and adjusting the `right` value so that there is only a short space between the text and the arrow - at the moment the arrow can be quite a long way to the right, which makes the text overall appear off-centre compared to the image above:

Screenshot 2023-06-30 at 10 31 26

You should also make sure that `flex-grow` and `flex-shrink` are set to 0 for the icon, or it will get squashed sometimes

After making those changes:
Screenshot 2023-06-30 at 10 33 33

tbx/static_src/sass/components/_client-block.scss
```diff
@@ -24,8 +24,11 @@
...
         @include media-query(large) {
+            grid-template-columns: repeat(4, 1fr);
+        }
+
+        @include media-query(menu-break-larger) {
             grid-template-columns: repeat(6, 1fr);
-            column-gap: 20px;
         }
 
         @include media-query(x-large) {
```
helenb, line 34: the column-gap of 130px is making the new version of the logos quite a bit smaller on my large monitor than they were on the old version, and also making the text wrap. I think we need to reduce that a bit.

Logos before:
Screenshot 2023-06-30 at 10 17 41

Logos after:
Screenshot 2023-06-30 at 10 17 48

helenb (changes requested):
Thanks @albinazs just a couple of tweaks that I would suggest. Could you also test how this appears on the service page - at the moment we have two different templates for the client logos - client-block.html on the service page, and client-logo-block.html on the proposition page - the latter being the streamfield. These can be tidied up when we remove the service page.

albinazs:
thank you, @helenb - committed all the suggestions.
As for the service page, currently, there's no possibility to add a link to the Client's logo, so the changes in this PR didn't affect it.

helenb:
> thank you, @helenb - committed all the suggestions. As for the service page, currently, there's no possibility to add a link to the Client's logo, so the changes in this PR didn't affect it.

Thanks - that looks much better now. I think the service page logos will be affected because the changes to styling that they share - the new column gap makes them look a bit close together. I think we should get this deployed, and alert Lisa to the fact that it slightly alters the service page design - it can be an incentive to get rid of these pages for good!

helenb:
Approved - not sure how to mark as approved without starting a new review

helenb (approved):

PR #504, opened 2023-07-10
1770 Add homepage hero images backend

victoriachan:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/views/4019870/pulses/1222851770)

- Add a new hero images (min 1, max 6) field to homepage. Note that this will need more FE work.
- Remove some unused homepage related models (HomePageHero and HomePageClient) to tidy up.

FE update:
1) Each time a user refreshes the homepage, they see a random hero image (new JS component adds `--visible` class to a random image from the list of hero images) 
2) Wagtail crop fill - desktop used in this case
3) Old unnecessary code tidied up
image

### How to Test

- Go to homepage admin, e.g. http://localhost:8000/admin/pages/5/edit/ There should now be a new 'Hero images' field that allows up to 6 images.

FE update:
- Refresh the page several times to see random hero image loaded each time

### MR Checklist

- [x] Consider adding unit tests, especially for bug fixes. If you don't, tell us why.
- [x] Consider updating documentation. If you don't, tell us why.
- [x] If relevant, list the environments / browsers in which you tested your changes.


tbx/project_styleguide/templates/patterns/pages/home/home_page.html
```diff
@@ -23,20 +23,19 @@ <h1 class="home__heading home__heading--desktop"><span>{{ page.hero_intro_primar
...
+                {# Shards SVG window + images #}
+                <div class="shards" data-shards-hero data-imagetotal="{{ page.hero_images.all|length }}">
+                    <div class="shards__wrapper">
+                        <svg class="shards__window" aria-hidden="true">
+                            <use class="shards__window--desktop" xlink:href="#shards-window-desktop"/>
+                            <use class="shards__window--mobile" xlink:href="#shards-window-mobile"/>
+                        </svg>
+                        <div data-shards-images>
+                            {% for hero_image in page.hero_images.all %}
+                                {% image hero_image.image fill-970x576 as hero_image %}
```
helenb, line 35: Because we are using the same image for desktop and mobile, you need to figure out the largest size that the image displays. This is actually at the 1022px breakpoint - just before switching to desktop. The div with the image fills the whole 1022px width so you'll need to update your fill- crop to use that width, and whatever height seems appropriate for the space it fills.

tbx/project_styleguide/templates/patterns/pages/home/home_page.html
```diff
@@ -23,20 +23,19 @@ <h1 class="home__heading home__heading--desktop"><span>{{ page.hero_intro_primar
...
+                                <div class="shards__image" style="background-image:url({{ hero_image.url }})"></div>
```
helenb, line 36: We could also set the background-position inline here based on the wagtail focal point, to ensure the CSS crop follows the wagtail focal-point - see https://docs.wagtail.org/en/latest/advanced_topics/images/focal_points.html

tbx/static_src/sass/components/_shards.scss
```diff
@@ -32,32 +44,21 @@
         }
     }
 
-    &__video {
+    &__image {
         position: absolute;
-        right: -60px;
-        top: 0;
-        max-width: 1024px;
-        max-height: 576px;
```
helenb, line 40: I think you still need a max-height here - the image gets very tall at about 860px breakpoint - compare to the live torchbox.com site at that breakpoint

helenb (commented):
Thanks @albinazs this is looking great - just a few things to address

tbx/project_styleguide/templates/patterns/pages/home/home_page.html
helenb: Update: Albina and I tried this on a call but it didn't look good because of the positioning of the shards, so we left it.

tbx/static_src/sass/components/_shards.scss
helenb: I wasn't able to repeat this on a second test, so we left this one too.

tbx/core/models.py
```diff
@@ -198,6 +157,20 @@ class HomePageFeaturedPost(Orderable):
...
+class HomePageHeroImage(Orderable):
+    page = ParentalKey(
+        "torchbox.HomePage", on_delete=models.CASCADE, related_name="hero_images"
+    )
+    image = models.ForeignKey(
+        "images.CustomImage",
+        help_text="The hero images will be displayed in a random order.",
+        null=True,
+        blank=False,
+        on_delete=models.SET_NULL,
```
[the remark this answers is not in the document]
helenb: I think we need at least one image for the template, so can we add validation to ensure each one is saved? (apologies if I've misunderstood)

engineervix (approved):

engineervix:
> Thanks for your feedback on this @engineervix . I discussed this with @victoriachan before we did this implementation, and we agreed that as there are only a limited number of images, creating a rendition for each one is not a massive overhead. The front-end implemenetation ensures that only the images needed are actually loaded by the browser. But I'll let @victoriachan consider your suggestion and feed back too.
> 
> @albinazs on that note, it would be great if you could add some comments in the HTML and CSS to explain that choice, so that someone doesn't accidentally make a change that will load all the images in the future.

Awesome, thank you!

tbx/core/models.py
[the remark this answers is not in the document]
helenb: Thanks both for explaining. I think probably setting the default image is overkill. What we will need for deployment is to make sure Lisa has lined up one image ready - uploaded and in wagtail - that we can select quickly after deploying. We could also potentially check when a lower traffic time to the site is for deployment.

PR #509, opened 2023-07-27
1389 | Home page HCM

chris-lawton:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/pulses/1230241389)

### Description of Changes Made
- Adds visual indicator to primary links and blog posts in HCM
- Add hover/focus state to the same links in HCM
- Ensure the external link icon can be seen in light or dark mode in HCM

### How to Test
- Visit the homepage locally
- Enable `forced-colors: active` for HCM and `prefers-color-scheme` for light/dark mode

### Screenshots

<details>
  <summary>Expand to see more</summary>

Screenshot 2023-07-27 at 11 43 08

Screenshot 2023-07-27 at 11 43 16

</details>

### MR Checklist

- [x] Add a description of your pull request and instructions for the reviewer to verify your work.
- [x] If your pull request is for a specific ticket, link to it in the description.
- [x] Stay on point and keep it small so the merge request can be easily reviewed.
- [ ] Consider adding unit tests, especially for bug fixes. If you don't, tell us why.
- [ ] Tests and linting passes.
- [ ] Consider updating documentation. If you don't, tell us why.
- [ ] If relevant, list the environments / browsers in which you tested your changes.

PR #513, opened 2023-07-28
8268 | Mobile header accessibility

chris-lawton:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/pulses/1193048268)

- Add and toggle `aria-expanded` attr on the mobile menu toggle
- Add skip link to the homepage
- Ensure hidden mobile menu links cannot receive focus
- Remove 100% height from logo as it was preventing links further down the page from being clicked on smaller viewports
- Move the mobile menu toggle button on the home page so it's in a logical tabbing order

### How to Test
- Visit the homepage locally and double check the above

### MR Checklist

### DIFF
--- a/tbx/static_src/sass/components/_nav-item.scss
+++ b/tbx/static_src/sass/components/_nav-item.scss
@@ -140,6 +140,11 @@
     // Mobile menu
     .mobile-nav & {
         border-top: 1px solid hsla(0, 0%, 100%, 0.2);
+        visibility: hidden;
+
+        .menu-active & {
+            visibility: visible;
+        }
 
         &--with-badge {
             display: flex;
--- a/tbx/project_styleguide/templates/patterns/atoms/menu-button/menu-button.html
+++ b/tbx/project_styleguide/templates/patterns/atoms/menu-button/menu-button.html
@@ -1,4 +1,4 @@
-<button class="menu-button" aria-label="Menu toggle" data-menu-toggle data-active="menu">
+<button type="button" class="menu-button" aria-label="Menu toggle" data-active="menu" aria-expanded="false" data-menu-toggle>
     <span></span>
     <span></span>
     <span></span>
--- a/tbx/project_styleguide/templates/patterns/base_page.html
+++ b/tbx/project_styleguide/templates/patterns/base_page.html
@@ -108,7 +108,7 @@
 
 {% block header %}
     {% include "patterns/atoms/skip-link/skip-link.html" %}
-    {% include "patterns/organisms/header/header.html" %}
+    {% include "patterns/organisms/header/header.html" with show_primary_nav=True %}
     {% include "patterns/atoms/menu-button/menu-button.html" %}
     {% include "patterns/molecules/navigation/mobile-nav.html" %}
 {% endblock header %}
--- a/tbx/project_styleguide/templates/patterns/organisms/header/header.html
+++ b/tbx/project_styleguide/templates/patterns/organisms/header/header.html
@@ -2,10 +2,12 @@
 <header class="header">
     <div class="header__inner">
         {% include "patterns/atoms/logo/logo.html" %}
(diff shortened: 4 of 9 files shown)


tbx/static_src/sass/components/_nav-item.scss
```diff
@@ -140,6 +140,11 @@
     // Mobile menu
     .mobile-nav & {
         border-top: 1px solid hsla(0, 0%, 100%, 0.2);
+        visibility: hidden;
```
helenb, line 143: Something is not quite right with the links in the menu drop-down on the home page, where the focus is skipping straight past the drop down. I don't think it is this rule that is the issue - I haven't tracked it down from a quick look.

helenb:
Everything looks good apart from an issue with tabbing the menu links in the dropdown on the home page. It would be nice if possible to also fix the 'more' button on the home page which currently can't be expanded via the keyboard.

tbx/static_src/sass/components/_nav-item.scss
chris-lawton: Have you got a video?

helenb: Here you go:

https://github.com/torchbox/wagtail-torchbox/assets/771869/534a0cf1-3512-4b70-a7aa-52121ef6944f

helenb:
> > It would be nice if possible to also fix the 'more' button on the home page which currently can't be expanded via the keyboard.
> 
> Should be fixed in #509. I think it can be opened via keyboard but there's no focus state so you can't tell it's the active element.

Great thanks - you're right it's just the focus state that was the issue

tbx/static_src/sass/components/_nav-item.scss
helenb: Flagging that this is still an issue although this is on staging now. I'll mention in stand-up.

chris-lawton: I think it's to do with the order of the DOM elements on the home page template, as with other pages on staging the focus moves in a logical order - https://torchbox.staging.torchbox.com/digital-products/

On the homepage template I moved `patterns/atoms/menu-button/menu-button.html` [down to here](https://github.com/torchbox/wagtail-torchbox/blob/staging/tbx/project_styleguide/templates/patterns/pages/home/home_page.html#L21) as otherwise the menu button was out of tab order - you can test this on https://torchbox.com/.

I think the fix will be to move `patterns/molecules/navigation/mobile-nav.html` to directly after `menu-button.html` on the homepage template I linked to above.

chris-lawton: Had more of a look into this and think i've resolved it here - https://github.com/torchbox/wagtail-torchbox/pull/513/commits/08590fdf0f0760b85e582636cadf95fca9e145e4

It looks like we were overriding the `{% block header %}` tag on `home_page.html` as we don't include the primary nav on it but moving elements out of that header messes up the tabbing order. 

So i've now matched the contents of `{% block header %}` on `home_page.html` to that of `base_page.html` but add a conditional on the primary nav as to wether to include it or not.

This in turn also fixes inconsistencies in logo placement when comparing the homepage to any other page.

helenb: Well done - that sounds like a good solution. I haven't been able to test it as for some reason that commit you linked to is not associated with a branch - not sure how that is possible?!

Screenshot 2023-08-15 at 07 51 39

chris-lawton: I rebased the branch shortly after which is probably why.

It's here now (https://github.com/torchbox/wagtail-torchbox/pull/513/commits/140b626ff4cdb2793e4de08a6076b152b2261754) but prob easier to checkout the branch and pull : )

helenb: Ah that would do it - this is working nicely now - well done. I will update staging.

PR #535, opened 2023-08-24
Tweaks to the sticky nav breakpoints on the report page - refactor

helenb:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/views/4019870)

Olly wanted the sticky navigation not to wrap on to two lines once the desktop styles (with the title to the left) are reached. We have to add a custom breakpoint that works with the final content on the live site which is as per this screenshot:

Screenshot 2023-08-24 at 10 41 09

With the existing re-use of the `in-page-nav` styles, this would have meant first resetting the styles for the large breakpoint, then for the custom breakpoint - for the `--report` modifier only. We are already overriding the styles for the unstuck nav, so it was starting to get quite complex, and would be hard to adjust again in the future - we're likely to have different content next year. So I have split out the report sticky navigation styling into its own component - `report-in-page-nav` so that the overrides are clearer. I've made it easy to update the custom breakpoint in one place if we need to do so in the future.

### How to Test
Create an impact report page locally, and adjust the content of the headings as per the draft page on the live site (screenshot above)

### Screenshots

<details>

Current styles on staging below 1280px

Screenshot 2023-08-24 at 10 45 34

New styles below 1280px

Screenshot 2023-08-24 at 10 46 09

</details>

### MR Checklist

- [x] If relevant, list the environments / browsers in which you tested your changes. - chrome on mac.

### DIFF
--- a/tbx/static_src/javascript/components/in-page-nav.js
+++ b/tbx/static_src/javascript/components/in-page-nav.js
@@ -9,7 +9,6 @@ class InPageNav {
         this.allSections = document.querySelectorAll('[data-service-section]');
         this.allMenuLinks = this.node.querySelectorAll('[data-in-page-nav] a');
         this.sentinelEl = document.getElementById('sentinel');
-        this.stickyClass = 'in-page-nav--stuck';
 
         this.initObserving();
         this.bindEvents();
--- a/tbx/static_src/sass/abstracts/_mixins.scss
+++ b/tbx/static_src/sass/abstracts/_mixins.scss
@@ -16,13 +16,6 @@
     }
 }
 
-// for one off media queries
-@mixin custom-media-query($query) {
-    @media only screen and #{$query} {
-        @content;
-    }
-}
-
 /* ============================================
     z-index
 */
--- a/tbx/static_src/sass/abstracts/_mixins.scss
+++ b/tbx/static_src/sass/abstracts/_mixins.scss
@@ -16,6 +16,13 @@
     }
 }
 
+// for one off media queries
+@mixin custom-media-query($query) {
+    @media only screen and #{$query} {
+        @content;
+    }
+}
+
 /* ============================================
     z-index
 */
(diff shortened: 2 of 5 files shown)


tbx/static_src/javascript/components/in-page-nav.js
```diff
@@ -9,7 +9,6 @@ class InPageNav {
         this.allSections = document.querySelectorAll('[data-service-section]');
         this.allMenuLinks = this.node.querySelectorAll('[data-in-page-nav] a');
         this.sentinelEl = document.getElementById('sentinel');
-        this.stickyClass = 'in-page-nav--stuck';
```
helenb, line 12: This was a random fix I spotted when debugging another issue my change had caused

tbx/static_src/sass/abstracts/_mixins.scss
```diff
@@ -16,6 +16,13 @@
     }
 }
 
+// for one off media queries
```
siimonevans, line 19: It's a nitpick, but I'm not convinced we need a mixin for this and it may encourage arbitrary media query values in future.

If it's intended for one-off purposes I'd just inline the media query as required (since the declaration is standard and a one liner).

If it's used frequently enough - as it appears it may be - then I think that justifies the addition of a global breakpoint in the config. Perhaps premature optimisation, but making the breakpoint value available elsewhere may be useful in future.

Your call! Not a big deal either way.

siimonevans (approved):

tbx/static_src/javascript/components/in-page-nav.js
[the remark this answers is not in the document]
helenb: No it's not referred to anywhere else in the file.

tbx/static_src/sass/abstracts/_mixins.scss
helenb: I've kept it custom in the file for now, as it was very specific for the content of the nav and is likely to change - so we don't want it used elsewhere. I've removed the custom media query mixin though - I was in two minds about whether that was a good idea.

PR #516, opened 2023-07-31
New StreamField-powered SubPropositionPage based on SubServicePage

engineervix:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/pulses/1193045284)

#### Overview

This PR introduces a new page type — `SubPropositionPage`, based on the existing `SubServicePage`, which has a fixed ordering of content on the rendered page. The  `SubPropositionPage` reorganizes the page content to use Streamfield blocks, thereby allowing content editors the freedom to arrange the content in any order.

#### More details

- Sub-service pages have 5 major sections:
  - Key Points
  - Testimonials
  - Processes
  - Work (Case Studies)
  - Thinking (Featured Blog Posts)
- For each of the above sections, I have created a new `StructBlock` with its own template
-  A new `StreamBlock` puts all the above blocks together, also including the blocks in `tbx.core.blocks.PageSectionStoryBlock`, which is part of the existing `SubServicePage`. 
  - these blocks are distinguished by grouping them into two groups: **Page sections** and **Content blocks**
  - **Content blocks** are 'standard' StreamField blocks which are enclosed in a `div` with class `streamfield`. However, when you add the **Page sections** blocks inside the same div, the styling breaks. Therefore, I created a new template for the new `StreamBlock`, with logic to ensure we don't place **Page sections** inside `<div class="streamfield">`. See 4fd8146a for details.
- The existing `patterns/molecules/hero/hero.html` template has an include for `patterns/molecules/in-page-nav/in-page-nav.html`.
  - A variable named `item` is passed to the included template, with the value of the `page` variable from the current context assigned to it. This `page` is expected to have certain fields/attributes, such as `process_section_title`, `testimonials_section_title`, etc. However, our new `SubPropositionPage` has been stripped of many of these fields. Moreover, the ordering of the fields in the in-page nav template is fixed, which would not work well for our use-case, where content editors can freely order the page sections.
  - Therefore, I have created new templates `patterns/molecules/hero/hero--sub-proposition.html` and `patterns/molecules/in-page-nav/in-page-nav--sub-proposition.html` to solve the above problem. See 52d66ac8 for details.
- I have made small changes (1c9b1234) to `patterns/organisms/client-block/client-block.html` and `patterns/organisms/latest-posts/latest-posts.html` to handle two situations where certain `page` attributes are not available. It seemed more appropriate in this case to just modify the templates to include this logic than to create new templates.
[...]

PR #537, opened 2023-08-29
Tidy up client logo block styling

helenb:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/views/4019870)

Previously, we had 2 templates for client logos, that both shared the same classes and styling:

```
/organisms/client-block/client-block.html
/molecules/streamfield/blocks/client-logo-block.html
```

In https://github.com/torchbox/wagtail-torchbox/pull/516 Victor has rationalised these to just use `/molecules/streamfield/blocks/client-logo-block.html` in the new proposition and sub-proposition page types.

The way we distinguished the two blocks previously was by whether they were within a streamfield block or not - now they always will be. The code therefore no longer needs so many overrides for the streamfield styling - they can move to the base styling apart from where we overriding streamfield styles set for the whole site.

Note that this change does break some of the styling on service and sub-service page types - however, at the point that this will be deployed we'll be un-publishing all of those old pages.

### How to Test

- Check out the branch and run the migrations (you may want to use `dellar-snapshot` to create a snapshot of the db on the `main` branch first, as staging now also has these migrations).
- Add a client logo block both on the proposition page and on the sub-proposition page (on the sub-proposition page they are part of the testimonials block), and check the styling matches the live site: http://localhost:8000/digital-products/#our-clients

### Screenshots

<details>

Client logos on the sub-proposition page

Screenshot 2023-08-29 at 14 06 09

</details>

### MR Checklist

### DIFF
--- a/tbx/static_src/sass/components/_client-block.scss
+++ b/tbx/static_src/sass/components/_client-block.scss
@@ -4,9 +4,17 @@
     background-color: var(--color--white);
     padding-bottom: ($gutter * 5);
 
+    .service-section & {
+        // on sub-proposition pages client logos sit within a service-section container
+        // on proposition pages they sit within a proposition-section container, which has a left and right margin
+        // this rule adds a similar margin to that used by propostion-section, so the client-logos apppear the same on both page types
+        margin: 0 $variable-gutter--small;
+    }
+
     &__container {
         @include container();
-        padding: 90px 0 0;
+        margin: 0;
+        padding: 0;
     }
 
     &__block {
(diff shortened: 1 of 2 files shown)


chris-lawton:
Thanks Helen, this will keep it simpler going forward!

The only thing I spotted was a rogue border on the sub-prop pages when in high contrast mode. It looks like it's being inherited from a border bottom on `.client-item__link`.

Sub-proposition page on the right:

Screenshot 2023-08-29 at 15 05 12

helenb:
> The only thing I spotted was a rogue border on the sub-prop pages when in high contrast mode. It looks like it's being inherited from a border bottom on `.client-item__link`.

Good spot - I will check again in HCM and get that tidied up

PR #544, opened 2023-10-11
Support/wagtail 51

lparsons396:
**This upgrades Wagtail to 5.1**

Support ticket: https://torchbox.monday.com/boards/1130687432/pulses/1264168986
Wagtail 5.1 release notes: https://docs.wagtail.org/en/stable/releases/5.1.html#

### Changes: 
- Modeladmin is soon to be deprecated and so the [Wagtail_modeladmin package](https://github.com/wagtail-nest/wagtail-modeladmin) has been used to continue supporting its features

Note: The decision was made to use the [modeladmin package](https://github.com/wagtail-nest/wagtail-modeladmin) instead of migrating to Snippets


victoriachan (changes requested):
I know we agreed to just use the ModelAdmin package for other projects. However as this is our own project, I'm wondering if we can update it to use snippets.

It looks like the model admin is only used for the Taxonomy section, with just the Service model. It was added in 2019, and the customisation is mainly to display the usage count. Usage count is now part of Wagtail core. I think we can ditch the whole customisation and just add Service as a standard snippet. 

Or if have the time, we could also [add custom SnippetViewSet](https://docs.wagtail.org/en/stable/topics/snippets/customising.html), to load the Service admin page in the same URL as before, and perhaps displaying the usage count on the listing so it can be seen in one glance. It looks quite straightforward.

Another more compelling reason to ditch the existing modeladmin code is that the usage count is broken. E.g. click on a usage on this page `/admin/taxonomy/service/` I didn't spot this earlier as this is a rarely used feature.

victoriachan (changes requested):
Looks good! Well done for adding the snippet admin group to map the Taxonomy menu item.

Just need to remove the `get_admin_urls_for_registration()` method from the snippet view class as it is a model admin method. Also I included some small nice-to-have customisation for the Services admin listing.

When I tested locally, the usage count is always 0 for the Service snippets, somehow, though the Service type is used in some blog posts. Even after running `rebuild_references_index`, they are 0. I'm not sure why. We might have to troubleshoot this on a separate ticket if this feature is required. Just something to note. 

It shouldn't hold us back as the previous implementation is already broken and not showing usage anyway.

victoriachan (approved):
Looks good! Thank you for your work!

zerolab:
Any change we can get this in?

helenb:
On the board it is with Anna for internal testing

PR #546, opened 2023-10-20
Move people hr feed to an env var instead of using the jobs index page

helenb:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/views/3284773/pulses/1297908290)

Previously the people hr feed url was fetched from a field in the jobs index page. Because this page was unpublished and had no visible function, Anna deleted it - this could easily happen again with another editor as it is not at all clear that the page is needed for code. Instead, this MR switches to an env var.

### How to Test
- check out the branch
- Make sure that the job_count cache is deleted on your local build, or temporarily comment out the cache code
- Copy the `PEOPLEHR_FEED_URL` variable in `local.py.example` to `local.py`
- Run your local build, and check that the job count displays in the drop-down on any wagtail page

### Screenshots

<details>

Screenshot 2023-10-20 at 10 00 51
</details>

### MR Checklist

### DIFF
--- a/tbx/core/context_processors.py
+++ b/tbx/core/context_processors.py
@@ -22,7 +22,7 @@ def peoplehr_jobs_count(request):
     if not job_count:
         peoplehr_feed = PeopleHRFeed()
         job_count = peoplehr_feed.get_job_count(settings.PEOPLEHR_FEED_URL)
-    cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)
+        cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)
 
     return {"job_count": job_count}
 
--- a/tbx/core/context_processors.py
+++ b/tbx/core/context_processors.py
@@ -2,7 +2,6 @@ from django.conf import settings
 from django.core.cache import cache
 
 from tbx.core.api import PeopleHRFeed
-from tbx.core.models import JobIndexPage
 
 
 def fb_app_id(request):
@@ -21,13 +20,9 @@ def peoplehr_jobs_count(request):
     job_count = cache.get(CACHE_KEY)
 
     if not job_count:
-        job_index = JobIndexPage.objects.first()
-        if job_index:
-            peoplehr_feed = PeopleHRFeed()
-            job_count = peoplehr_feed.get_job_count(job_index.jobs_xml_feed)
-        else:
-            job_count = 0
-        cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)
+        peoplehr_feed = PeopleHRFeed()
+        job_count = peoplehr_feed.get_job_count(settings.PEOPLEHR_FEED_URL)
+    cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)
 
     return {"job_count": job_count}
 
--- a/tbx/settings/base.py
+++ b/tbx/settings/base.py
@@ -568,3 +568,5 @@ BIRDBATH_PROCESSORS = [
     "birdbath.processors.users.UserEmailAnonymiser",
     "birdbath.processors.users.UserPasswordAnonymiser",
 ]
+# people hr feed
+PEOPLEHR_FEED_URL = env.get("PEOPLEHR_FEED_URL"),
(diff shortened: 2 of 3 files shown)


helenb:
@engineervix could you do a code review please? I wondered if I ought to create a test for this - let me know if you have any advice on tests.

tbx/core/context_processors.py
```diff
@@ -21,13 +20,9 @@ def peoplehr_jobs_count(request):
...
-        job_index = JobIndexPage.objects.first()
-        if job_index:
-            peoplehr_feed = PeopleHRFeed()
-            job_count = peoplehr_feed.get_job_count(job_index.jobs_xml_feed)
-        else:
-            job_count = 0
-        cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)
+        peoplehr_feed = PeopleHRFeed()
+        job_count = peoplehr_feed.get_job_count(settings.PEOPLEHR_FEED_URL)
+    cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)
```
engineervix, line 25: **Question**: Looking at the initial code, it seems the cache with key "job_count" was set if it didn't exist. Is there a reason why this has changed here so that `cache.set(CACHE_KEY, job_count, CACHE_TIMEOUT)` always runs?

tbx/settings/base.py
```diff
@@ -568,3 +568,5 @@
     "birdbath.processors.users.UserEmailAnonymiser",
     "birdbath.processors.users.UserPasswordAnonymiser",
 ]
+# people hr feed
+PEOPLEHR_FEED_URL = (env.get("PEOPLEHR_FEED_URL"),)
```
engineervix, line 572: **Suggestion**: I'd suggest specifying say an empty string as a default value, so that we don't get an error when `PEOPLEHR_FEED_URL` is not set.

engineervix (commented):
Thanks @helenb this looks good. Just one question and suggestion.

I considered using [Wagtail Settings](https://docs.wagtail.org/en/stable/reference/contrib/settings.html) for this but I think the environment variable approach is more appropriate, because we don't expect the URL to change.

Regarding tests, if you have time it would be great to have some unit tests. The way I'd go about testing the `peoplehr_jobs_count` function is as follows

- In the `setUp` method of the `TestCase`, create a [`RequestFactory`](https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#the-request-factory) instance, which will be passed as an argument to the `peoplehr_jobs_count` function when testing
- Create a 'fake' XML string with a structure similar to what we get from the PeopleHR Feed
- Use [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) to mock [requests](https://requests.readthedocs.io/) in `tbx.core.api` so that it always returns the XML string above
- Test that:
  - as long as the env variable is set, the job count returned matches the number of job entries in the fake XML string
  - before the first test, the cache entry with key `"job_count"` doesn't exist, and that afterwards, it is set with the value being the job count above.
  - if the environment variable is not set, the job count will be zero

helenb:
> Thanks @helenb this looks good. Just one question and suggestion.
> 
> I considered using [Wagtail Settings](https://docs.wagtail.org/en/stable/reference/contrib/settings.html) for this but I think the environment variable approach is more appropriate, because we don't expect the URL to change.
> 
> Regarding tests, if you have time it would be great to have some unit tests. The way I'd go about testing the `peoplehr_jobs_count` function is as follows
> 
> * In the `setUp` method of the `TestCase`, create a [`RequestFactory`](https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#the-request-factory) instance, which will be passed as an argument to the `peoplehr_jobs_count` function when testing
> * Create a 'fake' XML string with a structure similar to what we get from the PeopleHR Feed
> * Use [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) to mock [requests](https://requests.readthedocs.io/) in `tbx.core.api` so that it always returns the XML string above
> * Test that:
>   
>   * as long as the env variable is set, the job count returned matches the number of job entries in the fake XML string
>   * before the first test, the cache entry with key `"job_count"` doesn't exist, and that afterwards, it is set with the value of the job count above.
>   * if the environment variable is not set, the job count will be zero

thank you very much for the advice on this - I will have a go.

tbx/core/context_processors.py
helenb: that was a mistake, good spot

PR #548, opened 2023-11-02
Remove Service & SubService Pages

engineervix:
[Link to Ticket](https://torchbox.monday.com/boards/1192293412/pulses/1220423226)

With the change from ServicePages to PropositionPages, and SubServicePages to SubPropositionPages, we no longer need the models in the `services` app. This MR is to remove them from the codebase.

**Caveats**

- Unfortunately, we cannot completely delete the `services` app, as it contains migrations that are depended upon by other apps. This means we cannot also remove it from the `INSTALLED_APPS` list. 

### How to Test

- when you run migrations against `staging` / `production` / a fresh database, there should be no errors.
- in Wagtail admin, when creating a new Page, there should be no option to add a ServicePage or SubServicePage  

### MR Checklist

- [x] Consider updating documentation. If you don't, tell us why.
- [x] If relevant, list the environments / browsers in which you tested your changes.


tbx/propositions/migrations/0016_delete_subservicepagetosubpropositionpagemigration.py
```diff
@@ -0,0 +1,40 @@
...
+        ),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            state_operations=[],
+            database_operations=[
+                migrations.RunPython(delete_pages, nooperation),
+            ],
+        ),
```
victoriachan, line 35: This doesn't need to be in the `SeparateDatabaseAndState`.

```suggestion
        migrations.RunPython(delete_pages, migrations.RunPython.noop),
```

Also a trick I learnt is to use `migrations.RunPython.noop` for the reverse_code. This way you don't need to define an empty nooperation method above.

victoriachan (changes requested):
This is a tricky one! 

I didn't get error when I tested but because you used state_operations to delete the models, the tables are still lying around. 

I couldn't get DeleteModel to work normally here either due to all kinds of constraints. So I see why you used `state_operations`. I suspect because the db is already screwed up from historical reasons. 

So as long as we can get this tested thoroughly on staging, I suppose we are OK to go.

I just have a few small suggestions.

engineervix:
closing this PR as we'll be working on a new build

PR #555, opened 2023-11-14
Github Actions Enhancements

engineervix:
This MR adds a few enhancements to the CI configuration

- workflow to be triggered for **pushes** to any branch
- Split out the jobs and make use of caching to speed things up. If linting fails, no need to run tests.
- Use pre-commit action. This resulted in several templates being reformatted (`djhtml`). I have added the specific commits to `.git-blame-ignore-revs`
- Update Postgres version to match version specified in docker-compose.yml
- update actions to use more recent versions

> **Note**
>
> Many of the changes made are based on the [GitHub Actions config for the wagtail.org](https://github.com/wagtail/wagtail.org/blob/main/.github/workflows/ci.yml) project.

### MR Checklist

- [ ] If your pull request is for a specific ticket, link to it in the description.


victoriachan (approved):
Looks great! Thank you for the improvement!

.github/workflows/test.yml
```diff
@@ -1,11 +1,57 @@
 name: CI
 
 on:
-  push:
   pull_request:
+    branches: ['main']
```
helenb, line 5: I'm only just getting familiar with github actions, but I found in a personal project that I only needed to run the CI on `push`. There was no need to use `pull_request` at all. So say I had a feature branch, `feature/change-something` - CI would run when I push that branch. I then create a PR to merge `feature/change-something` to `main`. CI doesn't run again at the point I create my PR, but the PR shows the latest CI run for the branch. At the point that I hit 'merge', the `main` branch is pushed, and therefore the CI runs for the `main` branch. I'm interested to know how your setup works differently to that. I guess that CI wouldn't run each time you push a feature branch, only at the point you create a PR to main?

.github/workflows/test.yml
```diff
@@ -1,11 +1,57 @@
...
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('**/.nvmrc') }}-${{ hashFiles('**/package-lock.json') }}
+      - if: steps.node-cache.outputs.cache-hit != 'true'
+        run: npm ci --no-audit
+      - uses: actions/cache@v3
+        with:
+          path: ${{ env.HOME }}/.cache/pre-commit
+          key: ${{ runner.os }}-pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
+      - uses: pre-commit/action@v3.0.0
```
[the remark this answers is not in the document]
helenb: Query: I'm not familiar with pre-commit. Does this now rely on everyone to use pre-commit before pushing because it relies on cached files from running pre-commit? I think we'd want to avoid that as not everyone uses it, and as I understand it we've always kept use of it optional - we might want to check in and ask about that in the #tech slack channel.

.github/workflows/test.yml
```diff
@@ -1,11 +1,57 @@
...
+  test_js:
```
helenb, line 36: Nice work to split out the jobs

.git-blame-ignore-revs
```diff
@@ -0,0 +1,13 @@
+# git-blame ignored revisions
+# To configure, run
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+# Requires Git > 2.23
+# See https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt
+
+# Fix python formatting with black
+# pre-commit run black --color=always --all-files
+e6271cc9f60dcae600b011ab8e551bf1586bdc4d
```
helenb, line 9: Query: could we run black with the `--exclude migrations/` option to avoid having to reformat all the migration files?

helenb (commented):
Thank you Victor for picking this up. I have a few queries for you on the changes.

.github/workflows/test.yml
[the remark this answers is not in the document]
engineervix: @helenb **this doesn't require having pre-commit locally**. The pre-commit action runs all the tasks in the pre-commit configuration (already on the project), which are (mostly) linting / formatting tasks:

1. black
2. isort
3. flake8
4. eslint
5. prettier
6. stylelint
7. detect-secrets
8. djhtml

The advantage of using the pre-commit action is that we don't have to define separate tasks for each of the above. The caching helps speed up the workflow by avoiding redundant operations each time the action is invoked.

If a developer doesn't have pre-commit on their machine, then they are responsible for ensuring that whatever they push is correctly formatted and passes linting.

.github/workflows/test.yml
engineervix: With this config, pushing to any branch other than `main` or `staging` will not trigger CI, unless on a PR that targets the main branch. Therefore, if one creates a feature branch `feature/change-something` and pushes it to GitHub, nothing happens. When a PR to merge `feature/change-something` to `main` is created, GitHub Actions will be triggered, and any subsequent pushes to that feature branch.

I think your approach sounds better, because sometimes we may need to create a PR that targets an integration branch, then this configuration may not work well. So perhaps revising it to

```
name: CI

on:
  push:
    paths-ignore: ['docs/**']
```

This way, CI will run on every push. If documentation-only changes are made, we don't need to run CI.

.git-blame-ignore-revs
engineervix: Good spot! I realise black is actually configured to ignore them, will update to respect this

.github/workflows/test.yml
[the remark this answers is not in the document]
helenb: `If a developer doesn't have pre-commit on their machine, then they are responsible for ensuring that whatever they push is correctly formatted and passes linting.` - as long as the checks will still run if the developer doesn't run pre-commit themselves this is fine.

.git-blame-ignore-revs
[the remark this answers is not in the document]
helenb: I also learned about this from reviewing this PR :) Although I guess if we don't change all the migration files with black now (by using `--exclude migrations/`) then we won't need it any more for this particular PR.

.github/workflows/test.yml
helenb: that sounds good to me.

helenb (approved):
Thanks @engineervix this all looks good to me now, so let's get it merged.

PR #556, opened 2023-11-23
Add and enable image optimizers

zerolab:
https://www.loom.com/share/8303f38f8b374bbcbafce9cbb36df2e3?sid=e976fc37-197b-4749-9857-2fde64ca351a

Builds on #544

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,7 +101,7 @@ USER root
 
 # Install `psql`, useful for `manage.py dbshell`
 RUN apt-get update --yes --quiet && apt-get install --yes --quiet --no-install-recommends \
-    postgresql-client \
+    postgresql-client jpegoptim pngquant gifsicle libjpeg-progs webp \
     && apt-get autoremove && rm -rf /var/lib/apt/lists/*
 
 # Restore user
--- a/tbx/settings/base.py
+++ b/tbx/settings/base.py
@@ -569,3 +569,5 @@ BIRDBATH_PROCESSORS = [
 ]
 # people hr feed
 PEOPLEHR_FEED_URL = env.get("PEOPLEHR_FEED_URL", "")
+
+WILLOW_OPTIMIZERS = True


Dockerfile
```diff
@@ -101,7 +101,7 @@ USER root
 
 # Install `psql`, useful for `manage.py dbshell`
 RUN apt-get update --yes --quiet && apt-get install --yes --quiet --no-install-recommends \
-    postgresql-client \
+    postgresql-client jpegoptim pngquant gifsicle libjpeg-progs webp \
```
Morsey187, line 104: **Question** 
> Willow supports using dedicated libraries to optimize images. Out of the box, Willow comes with optimizers for [gifsicle](https://www.lcdf.org/gifsicle/), [jpegoptim](https://www.kokkonen.net/tjko/projects.html), [optipng](https://optipng.sourceforge.net/), [pngquant](https://pngquant.org/) and [cwebp](https://developers.google.com/speed/webp/docs/cwebp).

Willow seems to come with these default optimizers when setting WILLOW_OPTIMIZERS to true, are we sure the [cwebp](https://github.com/wagtail/Willow/blob/9eff9bb35265d2f778d6b8557d3bd1a0c4f3f998/willow/optimizers/cwebp.py) optimizer is the same as `webp` since the naming is slightly different.

Happy to test locally otherwise.

Morsey187 (commented):
Thanks for this Dan, this LGTM, just one question around webp

Dockerfile
zerolab: will check if we could get away with only installing `libwebp-dev`, but this works:

```
# dan in ~/.../wagtail-torchbox on git:chore/image-optimzer-binaries ● [17:07:44]
~ ❯❯❯ fab sh
tbx@dd448dc78614:/app$ cwebp
Usage:

   cwebp [options] -q quality input.png -o output.webp

where quality is between 0 (poor) to 100 (very good).
Typical value is around 80.

Try -longhelp for an exhaustive list of advanced options.
tbx@dd448dc78614:/app$
```

helenb:
@zerolab or @Morsey187 could we get this merged with the 5.1 changes? I'm trying to tidy up open MRs prior to starting the new build.

zerolab:
This is ready, so go forth and merge

PR #557, opened 2023-11-29
Set `BASIC_AUTH_DISABLE_CONSUMING_AUTHORIZATION_HEADER = True`

engineervix:
Wagtail requires Authorization header to be present for the previews

See https://github.com/tm-kn/django-basic-auth-ip-whitelist#basic_auth_disable_consuming_authorization_header


helenb (approved):
Looks good!

